PR #17, opened 2021-01-13
Adding Pointing default in GetMemoryHandler

anuragprat1k:
# Description

Adding Pointing default in GetMemoryHandler. For things like "where is x", now other than returning the location of x, the locobot and craftassist agents will also point to it. Made the following changes as well
* moved  `append_new_task` to the base `DialogObject` class so that it can be shared by all dialog objects
* added logging to not suppress exception traces (helpful for debugging)
* changed the base agent location helper to return a point instead of a region. Moved the transformation of the point to the region to the minecraft agent since this requirement is agent specific. 
* Fixed Locobot Point task

## Type of change

- [ ] This is a unit test

## Type of requested review

## Before and After

# Testing

Manually tested the folloing on the locobot assistant 
* Point to the chair
* where is the chair

Will need someone to test it on minecraft. 

# Checklist:

### DIFF
--- a/base_agent/dialogue_objects/dialogue_object.py
+++ b/base_agent/dialogue_objects/dialogue_object.py
@@ -80,6 +80,16 @@ class DialogueObject(object):
             self.finished = True
         return self.finished
 
+    def append_new_task(self, cls, data=None):
+        # this is badly named, FIXME
+        # add a tick to avoid two tasks having same timestamp
+        self.memory.add_tick()
+        if data is None:
+            self.memory.task_stack_push(cls, chat_effect=True)
+        else:
+            task = cls(self.agent, data)
+            self.memory.task_stack_push(task, chat_effect=True)
+
     def __repr__(self):
         return str(type(self))
 
--- a/base_agent/dialogue_objects/location_helpers.py
+++ b/base_agent/dialogue_objects/location_helpers.py
@@ -82,19 +82,3 @@ class ReferenceLocationInterpreter:
         interpreter.memory.update_recent_entities(mems)
 
         return mems
-
-
-def interpret_point_target(interpreter, speaker, d):
-    if d.get("location") is None:
-        # TODO other facings
-        raise ErrorWithResponse("I am not sure where you want me to point")
-    # TODO: We might want to specifically check for BETWEEN/INSIDE, I'm not sure
-    # what the +1s are in the return value
-    mems = interpreter.subinterpret["reference_locations"](interpreter, speaker, d["location"])
-    #    mems = interpret_reference_location(interpreter, speaker, d["location"])
-    steps, reldir = interpret_relative_direction(interpreter.agent, d)
-    loc, _ = interpreter.subinterpret["specify_locations"](
-        interpreter, speaker, mems, steps, reldir
-    )
-    #    loc, _ = compute_locations(interpreter, speaker, mems, steps, reldir)
-    return (loc[0], loc[1], loc[2], loc[0], loc[1]+1, loc[2])
--- a/base_agent/dialogue_objects/location_helpers.py
+++ b/base_agent/dialogue_objects/location_helpers.py
@@ -97,4 +97,4 @@ def interpret_point_target(interpreter, speaker, d):
         interpreter, speaker, mems, steps, reldir
     )
     #    loc, _ = compute_locations(interpreter, speaker, mems, steps, reldir)
-    return (loc[0], loc[1] + 1, loc[2], loc[0], loc[1] + 1, loc[2])
+    return (loc[0], loc[1], loc[2])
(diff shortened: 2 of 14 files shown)


base_agent/dialogue_objects/dialogue_object.py
```diff
@@ -80,6 +80,16 @@ def check_finished(self):
             self.finished = True
         return self.finished
 
+    def append_new_task(self, cls, data=None):
```
kavyasrinet, line 83: For a fix in later PR, we need better naming of parameters of this function

base_agent/dialogue_objects/location_helpers.py
```diff
@@ -97,4 +97,4 @@ def interpret_point_target(interpreter, speaker, d):
         interpreter, speaker, mems, steps, reldir
     )
     #    loc, _ = compute_locations(interpreter, speaker, mems, steps, reldir)
-    return (loc[0], loc[1] + 1, loc[2], loc[0], loc[1] + 1, loc[2])
```
kavyasrinet, line 100: Just a heads up to check that this change doesn't break any other command that would call this.

kavyasrinet (approved):
LGTM with a reminder to test on Minecraft.
Can you create an issue for that please ?

craftassist/agent/tasks.py
```diff
@@ -140,14 +140,16 @@ class Point(Task):
...
     Examples::
 
-        >>> target = [0, 0, 0, 2, 3, 4]
+        >>> target = [0, 0, 0]
         >>> task_data = {"target": target}
         >>> point = Point(agent, task_data)
     """
 
     def __init__(self, agent, task_data):
         super(Point, self).__init__()
```
aszlam, line 149: I don't think this should work, what if point target was a region?

anuragprat1k: Changed it so that they are converted to a region inside this function and location helper only passes a point. Are their any other codepaths that might pass it a region?

aszlam (changes requested):
I think it would be better to continue to have point be a region in general; and on the locobot take the mean if a non-empty region is given (e.g. point targets are still minx, maxx, miny, maxy, minz, maxz).  I think agree its better to handle the y+1 as necessary on the craftassist side instead of the base agent location helper

craftassist/agent/tasks.py
aszlam: lol, there are several places where we are using the agent point directly, and probably should be putting a task instead:
https://github.com/facebookresearch/droidlet/blob/main/craftassist/agent/dialogue_objects/put_memory_handler.py#L126 
https://github.com/facebookresearch/droidlet/blob/main/base_agent/dialogue_objects/dialogue_object.py#L373

here: https://github.com/facebookresearch/droidlet/blob/main/craftassist/agent/dialogue_objects/mc_interpreter.py#L451
its using the Task and this code path.  

I agree with your off-github comments that we should clean this up

snyxan:
> I think it would be better to continue to have point be a region in general; and on the locobot take the mean if a non-empty region is given (e.g. point targets are still minx, maxx, miny, maxy, minz, maxz). I think agree its better to handle the y+1 as necessary on the craftassist side instead of the base agent location helper

Agreed.

craftassist/agent/tasks.py
anuragprat1k: Changed it back to a region.

anuragprat1k:
> I think it would be better to continue to have point be a region in general; and on the locobot take the mean if a non-empty region is given (e.g. point targets are still minx, maxx, miny, maxy, minz, maxz). I think agree its better to handle the y+1 as necessary on the craftassist side instead of the base agent location helper

Changed it back to point accepting a region and handling the region pointing specially (by just taking the min xyz) for locobot.

base_agent/dialogue_objects/location_helpers.py
```diff
@@ -97,4 +97,4 @@ def interpret_point_target(interpreter, speaker, d):
...
+    return (loc[0], loc[1], loc[2], loc[0], loc[1]+1, loc[2])
```
aszlam, line 100: am worried if there isn't a +1 on the first loc[1] its going to do weird things to mobs/players when it points (which in mc is flashing a block into and out of existence).

locobot/agent/tasks.py
```diff
@@ -77,15 +77,20 @@ def __init__(self, agent, task_data):
         super(Point, self).__init__()
         self.target = np.array(task_data["target"])
 
+    def get_pt_from_region(self, region):
+        assert len(region) == 6, "Region list has less than 6 elements (minx, miny, minz, maxx, maxy, maxz)"
+        return region[:3] # just return the min xyz for now
```
aszlam, line 82: take the mean, no?  probably doesn't matter that much

anuragprat1k: this is an arbitrary choice. Going to keep it as the first xyz since the coords for objects output by locobot perception is approximately the centre of the object.

anuragprat1k:
@aszlam pulled out the erstwhile `interpret_point_target` into its own subinterpreter called `PointTargetInterpreter` since the logic here is agent specific - in that there is this assumption of +1s for minecraft that we don't need for the locobot.

aszlam (approved):

PR #27, opened 2021-01-25
JSON schema for parser datasets

Rebecca-Qian:
# Description

This PR creates a data validator for all datasets used by the semantic parser, using JSON schema (version 7). Schema organization follows organization of the grammar spec in markdown.

Each schema file contains a unique ID (`$id`) used  to resolve cross references, description, a set of definitions of node types, and possible keys (properties). Note that subschemas here do not contain keys and are only sets of definitions.

For a given node, the schema defines its type, accepted children ("properties"), and constraints such as required properties or array length. Nodes can refer to nodes in other schemas, or themselves, eg.
```
"schematic": {
      "type": "object",
      "properties": {
        "triples": {
          "$ref": "filters.schema.json#/definitions/triples"
        },
        "repeat": {
          "$ref": "#/definitions/repeat"
        },
        "text_span": {
          "$ref": "#/definitions/span"
        }
      },
      "required": [
        "triples"
      ]
    },
```
Here, the definition for `triples` references the filters schema file, whereas `repeat` and `text_span` are defined in the same file.

We can encode logic for combining operations in several ways.
If-else:
```
      "if": {
        "properties": {
          "repeat_key": {
            "const": "FOR"
          }
        },
        "required": [
          "repeat_key"
        ]
      },
      "then": {
        "required": [
          "repeat_count"
        ]
      }
```
This means if the properties include `repeat_key` with value `FOR`, then the properties must include `repeat_count`.

One of:
```
"oneOf": [
    {
      "properties": {
        "dialogue_type": {
          "const": "HUMAN_GIVE_COMMAND"
        },
        "action_sequence": {
          "$ref": "#/definitions/action_sequence"
        }
      }
    },
    {
      "properties": {
        "dialogue_type": {
          "enum": [
            "NOOP",
            "BOT_CAPABILITIES"
          ]
        }
      }
    }]
```
Strictly one of these schemas are allowed, so `action_sequence` and `NOOP` commands are mutually exclusive. Oneof can be used at different levels, eg. specify sets of properties, values of nodes, or values of properties.

allOf and anyOf are similar, where anyOf matches against any number of schemas more than one.

Fixes # (issue)
Data validator is motivated by frequency of dataset bugs after codemods and grammar changes.

## Type of change

- [ x] New feature (non-breaking change that adds a new functionality)
- [ ] This is a unit test

## Type of requested review

- [ x] I want a thorough review of the implementation.

## Before and After
Ran the validator on parser training data and it has caught many bugs
`annotated.txt`: P165490927
- `location_type`
- `argval`
- inconsistent literals
`templated.txt`: P165616720
- `filters` missing and keys are in parent dict

# Testing
`python validate_json.py --data_path [FILE_TO_VALIDATE]`

# Checklist:

- [x ] I have performed manual end-to-end testing of the feature in my environment.
- [x ] I have added Docstrings and comments to the code.
- [x ] I have added relevant collaborators to review the PR before merge.


anuragprat1k (approved):

kavyasrinet (commented):
This PR doesn't have the markdown/HTML view so I suggest either keeping the json files under `base_agent/documents/json_schema` or adding the Markdown support. having the Also, `validate_json.py` doesn't belong here either. Perhaps somewhere more central so DialogueManager can use the same script ?

Can you also confirm whether the PR for injecting it in Dialogue Manager (wherever we use the logical forms, either from querying the model or fetching from ground truth) will follow ?
If yes, please add a note.

Rebecca-Qian:
Addressed comments
- Schemas moved to `base_agent/documents/json_schema/`
- `validate_json.py` now in `craftassist/test/`. This may look out of place right now, but over the half we'll be writing more and different kinds of tests, so I'm envisioning a test directory structure such as `test/agent/` for current unit tests, `test/integration/` and `test/dataset/`. cc @anuragprat1k @kavyasrinet 

Also refactored the validation script, since next step is to plug it into DM.

Next steps
- Integrate data validation into dialogue manager for logical forms going into interpreter
- Documentation: update markdown files, look into markdown/HTML conversion for JSON  
- Integrate validator into internal annotation tools, i.e. AutoComplete

We should start validating dataset changes and existing logical forms using the script now!

kavyasrinet (commented):
Fill command also needs a triple.

PR #58, opened 2021-01-29
clean up exception handling and only safe-handle whitelisted Exceptions

soumith:
# Description

The exception handling in `base_agent` was eating all fatal exceptions, even though the intention was to catch exceptions coming from modules with expected failures such as Dialog Parser / Interpreter of the form "I don't understand what you are trying to say" where there are just out-of-distribution issues in the NLP stack.

So, for example if the database failed or if the connection with the robot failed, the agent was like continuing as if nothing happened.

This cleans up stuff in a way that only whitelisted exceptions (currently only `ErrorWithResponse`) are eaten and everything else is re-raised.

A debug mode is allowed, called `DROIDLET_DEBUG_MODE=1` where all exceptions are eaten (i.e old behavior)

Fixes https://github.com/facebookresearch/droidlet/issues/54

## Type of change

## Type of requested review

## Before and After

# Testing

# Checklist:

### DIFF
--- a/base_agent/core.py
+++ b/base_agent/core.py
@@ -18,9 +18,6 @@ class BaseAgent:
                 self.step()
             except Exception as e:
                 self.handle_exception(e)
-                if os.getenv('DEVMODE'):
-                    # raise immediately in dev mode to catch and fix errors
-                    raise e
 
     def step(self):
         self.perceive()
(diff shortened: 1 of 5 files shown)


kavyasrinet (approved):
Thanks for adding this. LGTM.

PR #64, opened 2021-01-29
Specify package versions in requirements

snyxan:
# Description

We should explicitly specify compatible versions of packages in requirements. This works for craftassist. Since I can't run locobot right now, could you also add versions to `locobot/requirements.txt` @soumith @anuragprat1k ?

Fixes # (issue)

## Type of change

- [ ] This is a unit test

## Type of requested review

# Testing

Tests passed locally.

# Checklist:


requirements.txt
```diff
@@ -29,9 +29,9 @@ tqdm==4.41.1
 word2number==1.1
 transformers==3.3.1
 https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-2.0.0/en_core_web_sm-2.0.0.tar.gz
-coverage
-flask-cors
-flask
+coverage==5.3
+flask-cors==3.0.9
```
soumith, line 33: exact versions seem an overkill.

I prefer if you change it to something like `flask-cors>=3.0`

soumith (changes requested):
instead of exact versions, let's move to `>=` of a minimum major version.

This will allow less breakage when versions of packages upgrade underneath us

requirements.txt
snyxan: I was worried about package backwards compatibility issue but realized that we can rely on CI to detect it. So aggressively changed all package version req to `>=` except for `transformer` which requires a certain version.

soumith (approved):

snyxan:
@kavyasrinet the conflict comes from prettytable package you added. Could you specify a minimum version for that?

kavyasrinet:
> @kavyasrinet the conflict comes from prettytable package you added. Could you specify a minimum version for that?

Yeah just checked - `2.0.0`

PR #80, opened 2021-02-02
use a different function for shell script absolute path resolution to be compatible with OSX

soumith:
# Description

Made the data scripts compatible with OSX

Fixes https://github.com/facebookresearch/droidlet/issues/57

## Type of change

- [ ] This is a unit test

## Type of requested review

### DIFF
--- a/tools/data_scripts/compare_directory_hash.sh
+++ b/tools/data_scripts/compare_directory_hash.sh
@@ -2,7 +2,11 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 
 # This script checks if models and datasets are up to date, and either triggers a download or gives the user a warning to update local files.
-ROOTDIR=$(readlink -f $(dirname "$0")/../../)
+function pyabspath() {
+    python -c "import os; import sys; print(os.path.realpath(sys.argv[1]))" $1
+}
+
+ROOTDIR=$(pyabspath $(dirname "$0")/../../)
 echo "$ROOTDIR"
 
 if [ -z $1 ]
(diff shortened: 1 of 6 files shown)


anuragprat1k (approved):

tools/data_scripts/compare_directory_hash.sh
```diff
@@ -2,7 +2,11 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 
 # This script checks if models and datasets are up to date, and either triggers a download or gives the user a warning to update local files.
-ROOTDIR=$(readlink -f $(dirname "$0")/../../)
+function pyabspath() {
+    python -c "import os; import sys; print(os.path.realpath(sys.argv[1]))" $1
```
Rebecca-Qian, line 6: In bash this breaks if the argument contains spaces. This is an edge case, but if someone has a directory name with spaces, the script would fail. What do you think about changing the shebang to zsh for all data scripts?

Rebecca-Qian (approved):
Other than the bash args issue, LGTM.

tools/data_scripts/compare_directory_hash.sh
soumith: good point, i think it is easy to fix with single quotes, let me quickly check

soumith: it seems to work for me in bash on both Linux and OSX.

I took this snippet, and put it as `foo.sh` inside the directory `foo bar` (with a space)

```bash
function pyabspath() {
    python -c "import os; import sys; print(os.path.realpath(sys.argv[1]))" $1
}

ROOTDIR=$(pyabspath $(dirname "$0"))
echo $ROOTDIR
```

Then, I ran the script with `bash foo.sh` (from inside the `foo bar` directory).

Both on Linux and OSX, I got the outputs:

`/Users/soumith/foo bar` (OSX)
`/raid/soumith//foo bar` (Linux)

PR #88, opened 2021-02-03
Handling Pyro4 exceptions + sending ErrorWithResponse msg to the dashboard

anuragprat1k:
# Description

* Converting a frequent Pyro4 exception to our whitelisted exception (more details on that in https://github.com/facebookresearch/droidlet/pull/58) 
* Changing ErrorWithResponse handling to also pass the exception text back to the dashboard to make it more meaningful.

## Type of change

- [ ] This is a unit test

## Type of requested review

# Testing
Manually tested it
pyro4except

# Checklist:

### DIFF
--- a/base_agent/loco_mc_agent.py
+++ b/base_agent/loco_mc_agent.py
@@ -2,7 +2,7 @@
 Copyright (c) Facebook, Inc. and its affiliates.
 """
 import sys
-
+import os 
 import logging
 import os
 import random
@@ -197,7 +197,7 @@ class LocoMCAgent(BaseAgent):
         # if so, we raise a reasonable message to the user, and then do some clean
         # up and continue
         if isinstance(e, ErrorWithResponse):
-            self.send_chat("Oops! I got confused and wasn't able to complete my last task :(")
+            self.send_chat("Oops! Ran into an exception.\n'{}''".format(e.chat))
             self.memory.task_stack_clear()
             self.dialogue_manager.dialogue_stack.clear()
             self.uncaught_error_count += 1
--- a/locobot/agent/locobot_mover.py
+++ b/locobot/agent/locobot_mover.py
@@ -21,12 +21,12 @@ from locobot_mover_utils import (
     base_canonical_coords_to_pyrobot_coords,
     xyz_pyrobot_to_canonical_coords,
 )
-from tenacity import retry, stop_after_attempt
+from tenacity import retry, stop_after_attempt, wait_fixed
 
 Pyro4.config.SERIALIZER = "pickle"
 Pyro4.config.SERIALIZERS_ACCEPTED.add("pickle")
 
-@retry(stop=stop_after_attempt(5))
+@retry(reraise=True, stop=stop_after_attempt(5), wait=wait_fixed(0.5))
 def safe_call(f, *args):
     try:
         return f(*args)
--- a/locobot/agent/locobot_mover.py
+++ b/locobot/agent/locobot_mover.py
@@ -8,6 +8,7 @@ import cv2
 import os
 import math
 import copy
+from base_agent.base_util import ErrorWithResponse
 from perception import RGBDepth
 from objects import Marker, Pos
 from locobot_mover_utils import (
@@ -24,6 +25,12 @@ from locobot_mover_utils import (
 Pyro4.config.SERIALIZER = "pickle"
 Pyro4.config.SERIALIZERS_ACCEPTED.add("pickle")
 
+def safe_call(f, *args):
+    try:
+        return f(*args)
+    except Pyro4.errors.ConnectionClosedError as e:
+        msg = "{} - {}".format(f._RemoteMethod__name, e)
+        raise ErrorWithResponse(msg)
 
 class LoCoBotMover:
     """Implements methods that call the physical interfaces of the Locobot.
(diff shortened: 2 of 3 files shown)


locobot/agent/locobot_mover.py
```diff
@@ -38,16 +45,17 @@ def __init__(self, ip=None, backend="locobot"):
         self.close_loop = False if backend == "habitat" else True
         self.curr_look_dir = np.array([0, 0, 1])  # initial look dir is along the z-axis
 
-        intrinsic_mat = self.bot.get_intrinsics()
+        intrinsic_mat = safe_call(self.bot.get_intrinsics)
```
soumith, line 48: i'm not convinced that these `safe_call`s are actually future-safe.

Presumably, the Pyro4 connection can be closed at any point in time, not just in these specific `safe_call`s in this PR.
So,
1. Why did you only make these 3 callsites to be `safe_call` and leave the others as-is?
2. What happens if a `safe_call` fails while doing something critical or important? Should we actually have a `reset() / revert()` mechanism for the robot to undo or get out of a multi-step task that it gave up on?
3. Why not do a `retry_call` where it retries a few times with a `time.sleep` in between, and if none of it works, then actually just fatally crap out?

anuragprat1k: > i'm not convinced that these `safe_call`s are actually future-safe.
> 
> Presumably, the Pyro4 connection can be closed at any point in time, not just in these specific `safe_call`s in this PR.
> So,
> 
> 1. Why did you only make these 3 callsites to be `safe_call` and leave the others as-is?
> 2. What happens if a `safe_call` fails while doing something critical or important? Should we actually have a `reset() / revert()` mechanism for the robot to undo or get out of a multi-step task that it gave up on?
> 3. Why not do a `retry_call` where it retries a few times with a `time.sleep` in between, and if none of it works, then actually just fatally crap out?

Answers:
1. Can easily add `safe_call` to all callsites. Added this only to the constructor and `get_base_state` because these are the canaries in the coal mine in a way - they get called the first and the most times. Also, anecdotally haven't run into exceptions with the other APIs and I don't know of a reliable way to simulate their errors either (and hence don't know everything we'll need to (or not) catch).
2. Some of the Pyro4 errors succeed on retries. If they continue to fail fatally, the exception will be raised eventually based on the mechanism here https://github.com/facebookresearch/droidlet/blob/bf813047ed0e74479420217d47f8b990e43d5231/base_agent/loco_mc_agent.py#L163.
3. We can do that, but the design currently essentially achieves that (ie maybe succeeds on retried or fails fatally eventually). Don't think it is worth the investment now given that these errors are only sporadic.

anuragprat1k: Also, good call out about `reset() / revert()` and this does not fix that. Let's cross that bridge when we get there?

base_agent/loco_mc_agent.py
```diff
@@ -156,7 +156,7 @@ def handle_exception(self, e):
         # if so, we raise a reasonable message to the user, and then do some clean
         # up and continue
         if isinstance(e, ErrorWithResponse):
-            self.send_chat("Oops! I got confused and wasn't able to complete my last task :(")
+            self.send_chat("Oops! Ran into an exception.\n'{}''".format(e.chat))
```
kavyasrinet, line 159: overall, I like the idea of sending across detailed error messages to the dashboard

anuragprat1k:
@soumith @kavyasrinet could you please sign off or request changes? thanks!

locobot/agent/locobot_mover.py
soumith: on (2), the important distinction is whether they are retried before the current codepath / task is finished, or they are retried on the next execution.

"bot go to the door"

`safe_call` fails, and the bot to the door task fails.

"bot go to the door"

`retry_call` tries ~5 times with sleep(1), and passes, continues and finishes the task.

That's the important distinction / decision to deliberately make, whether the call will retry before giving up on the current codepath and raising an exception.

soumith:
i made another comment on `safe_call`, I think it is worth adding a blocking retry to `safe_call` so that things are just generally smoother. see what you think

locobot/agent/locobot_mover.py
anuragprat1k: That makes sense. Added retries with wait using https://github.com/jd/tenacity

anuragprat1k:
> i made another comment on `safe_call`, I think it is worth adding a blocking retry to `safe_call` so that things are just generally smoother. see what you think

Added a waiting retry to safe call.

soumith (approved):
looks good.

one thing I realized is `safe_call` shouldn't be used  yet in it's current state for bot movement commands.

With bot movement, we have to check if a "Connection Closed" can be generated by Pyro4 after sending the bot command, i.e. while the bot command is executing. If so, retrying will involve issuing the move command twice.

anuragprat1k:
> looks good.
> 
> one thing I realized is `safe_call` shouldn't be used yet in it's current state for bot movement commands.
> 
> With bot movement, we have to check if a "Connection Closed" can be generated by Pyro4 after sending the bot command, i.e. while the bot command is executing. If so, retrying will involve issuing the move command twice.

Yeah, haven't seen that happen anecdotally with movement commands but there's definitely lot more work with those commands (like having async commands, being able to interrupt commands etc)

PR #106, opened 2021-02-04
Pointing by moving closer

anuragprat1k:
# Description

* Changed the point task on the locobot to include a move - it moves close to the target and then points to it
* Cleaned up some pyrobot-droidlet coordindate conversions to use a rotation matrix
* Added tests for some transformation utils
* Test video for "point to the gray chair" - https://youtu.be/c82Fpn0C6XI 
* Increased the default test threshold for the detector 

Overall, this still is brittle because move is brittle - move without using [dslam](https://github.com/facebookresearch/droidlet/pull/77) means the errors from the detector dominate and the robot crashes into things, and move with using [dslam](https://github.com/facebookresearch/droidlet/pull/77) needs more testing to ensure reliable behavior.

## Type of change

- [ ] Refactor
- [ ] This is a unit test

## Type of requested review

# Testing

# Checklist:


locobot/agent/locobot_mover.py
```diff
@@ -155,8 +159,24 @@ def look_at(self, obj_pos, yaw_deg, pitch_deg):
...
+        Args:
+            target_pos ([x,y,z]): canonical coordinates to point to.
+
+        Returns:
+            string "finished"
+        """
+        # Part 1: move to within a certain distance of the target.
         pos = self.get_base_pos_in_canonical_coords()
         yaw_rad, pitch_rad = get_camera_angles([pos[0], ARM_HEIGHT, pos[1]], target_pos)
+        self.move_absolute([get_move_target_for_point(pos, target_pos, yaw_rad)])
```
aszlam, line 175: imo this should be in tasks.py:  a Point task should first place a Move task, instead of everything being in a blob in the mover

locobot/agent/locobot_mover_utils.py
```diff
@@ -63,11 +63,44 @@ def transform_pose(XYZ, current_pose):
     return XYZ
 
 
+def get_move_target_for_point(base_pos, target, yaw, eps=1):
+    """
```
aszlam, line 67: could we separate out the logic for where to move for the point (from the coordinate transforms) and put it in tasks.py?

locobot/agent/locobot_mover.py
anuragprat1k: Added a move task inside point.

locobot/agent/locobot_mover_utils.py
anuragprat1k: The util still lives here (will move this to a better named utils file in the medium term) but gets called in tasks.py. Does that work?

aszlam: yeah, as long as the tasks steps are controlled by the agent loop I am fine with merging.  I would rather the logic for choosing where to move is in an interpreter... why a util file?  will this logic be used somewhere else?

aszlam (approved):

PR #119, opened 2021-02-09
multitask and events core

aszlam:
# Description

This sets up multi-tasking and events.    Main change is that instead of using the time order to decide which tasks to .step(), tasks are stepped when their prio is >0.   Implementing this requires extra machinery for making sure that tasks that are supposed to be executed in a specific sequence get executed in the correct sequence.  Some associated changes: 

- The Loop task in locobot and mc is merged and generalized into a ControlBlock.  
- Tasks are added in the interpreter body, instead of in the action handlers
- added several new Conditions

## Type of change

- [x] Refactor
- [ ] This is a unit test

## Type of requested review

## Before and After

# Testing

# Checklist:


aszlam:
This is not actually ready to merge yet, I just wanted to add y'all and ask for a look before it gets too out of control.  In particular, undo still isn't done, the multiplexer isn't included (all active tasks are stepped) and there are not yet tests for some new features.  In addition, one of the old loop tests is flaky and needs to be rewritten, will also do that in the next few commits here

aszlam:
Also, in order to get events and multi-task for serious, will need to do a bit of grammar work and make some data... in particular, many of the places where we currently have "stop_condition" need to instead have "remove_condition" (or we should change the spec)

base_agent/base_memory_schema.sql
```diff
@@ -93,10 +93,19 @@ CREATE TABLE Programs (
 );
 
 
+
+-- prio > 0 :  run me if possible, check my stop condition
```
snyxan, line 97: is there a range for prio? like 1-10

base_agent/memory_nodes.py
```diff
@@ -789,17 +807,92 @@ def create(cls, memory, task: Task) -> str:
...
+        finished = -1 :  task has not completed
+        
+        this method updates these columns of the DB for each of the keys if have values
+        if force_db_update is set, these will be updated with the relevant attr from self.task
+        even if it is not in the status dict.  
+        if force_task_update is set, the information will go the other way,
+        and whatever is in the dict will be put on the task.
+        """
+        status_out = {}
+        for k in ["prio", "running", "finished"]:
```
snyxan, line 859: finished -> finished_at?

snyxan:
On high level the design looks good to me.

base_agent/sql_memory.py
```diff
@@ -762,10 +758,11 @@ def task_stack_update_task(self, memid: str, task: Task):
         Examples ::
             >>> task = Move(agent, {"target": pos_to_np([0, 12, 0]), "approx" : 3})
             >>> memid = '10517cc584844659907ccfa6161e9d32'
-            >>> task_stack_update_task(task, parent_memid)
+            >>> task_stack_update_task(task, memid)
         """
         self.db_write("UPDATE Tasks SET pickled=? WHERE uuid=?", self.safe_pickle(task), memid)
 
+    # TORCH this
     def task_stack_peek(self) -> Optional["TaskNode"]:
```
snyxan, line 766: prio will be used with multiplexer to determine which tasks to peek?

snyxan: in following commit

base_agent/sql_memory.py
```diff
@@ -29,7 +29,17 @@
     NODELIST,
 )
 
-
+NONPICKLE_ATTRS = [
```
anuragprat1k, line 32: why are these nonpickle-able?

base_agent/memory_nodes.py
```diff
@@ -789,17 +807,92 @@ def create(cls, memory, task: Task) -> str:
...
+    # use this to update prio or running, don't do it directly on task or in db!!
+    def get_update_status(self, status, force_db_update=True, force_task_update=True):
+        """
+        status is a dict with possible keys "prio", "running", "finished".
+        
+        prio > 0 :  run me if possible, check my stop condition
+        prio = 0 :  check my start condition, run if true
+        prio < 0 :  don't even check my start condition
+
+        running = 1 :  task should be stepped if possible and not explicitly paused
```
anuragprat1k, line 846: how is the task stepped in this function?

base_agent/memory_nodes.py
```diff
@@ -789,17 +807,92 @@ def create(cls, memory, task: Task) -> str:
...
+
+    def update_task(self, task=None):
+        task = task or self.task
+        self.memory.db_write(
+            "UPDATE Tasks SET pickled=? WHERE uuid=?", self.memory.safe_pickle(task), self.memid
+        )
+
+    # FIXME TODO don't need paused, set prio to 0 and have a condtion for unpausing
+    # use this to update prio or running, don't do it directly on task or in db!!
+    def get_update_status(self, status, force_db_update=True, force_task_update=True):
```
anuragprat1k, line 838: Is this name reflective of what the function does? From the name it seems like the function just does read operations (gets statues) but from the docstring it looks like the function does a lot of updates as well.

base_agent/sql_memory.py
```diff
@@ -1120,37 +1066,32 @@ def dump(self, sql_file, dict_memory_file=None):
         sql_file.write("\n".join(self.db.iterdump()))
         if dict_memory_file is not None:
             import io
-            import pickle
 
             assert type(dict_memory_file) == io.BufferedWriter
             dict_memory = {"task_db": self.task_db}
             pickle.dump(dict_memory, dict_memory_file)
 
+    def reinstate_attrs(self, obj):
```
anuragprat1k, line 1074: add docstring for this function.

base_agent/task.py
```diff
@@ -28,56 +25,154 @@ class Task(object):
...
+        TaskNode(agent.memory, self.memid).update_task(task=self)
 
-    def step(self, agent):
-        """The actual execution of a single step of the task is defined here."""
-        # todo? make it so something stopped by condition can be resumed?
-        if self.stop_condition.check():
-            self.finished = True
-            return
-        return
+    def check_remove_and_running_children(stepfn):
```
anuragprat1k, line 46: docstring here as well. Why is this just a shell for `modified_step`?

base_agent/task.py
```diff
@@ -28,56 +25,154 @@ class Task(object):
...
+
+    check_remove_and_running_children = staticmethod(check_remove_and_running_children)
+
+
+# FIXME new_tasks_fn --> new_tasks
+# FIXME/TODO: name any nonpicklable attributes in the object
+# should we just have this be the behavior of the step of an unsublcassed Task object?
+class ControlBlock(Task):
+    """Container for task control
+
```
anuragprat1k, line 100: More details here since this is a major change. What is the scope of the ControlBlock?

base_agent/task.py
```diff
@@ -28,56 +25,154 @@ class Task(object):
...
+
+# the extra complexity here is that we are
+# allowing loops that are finite sequences, and loops
+# driven by a task generator, AND allowing child tasks to be
+# appended to a parent after the parent is placed in memory or is active
+# Probably can clean this up
+# without building this mini-stack,
+# just by using stop/remove conditions.... TODO?
+class TaskGenerator:
```
anuragprat1k, line 137: Docstring here as well - what is TaskGenerator?

base_agent/condition.py
```diff
@@ -52,6 +73,33 @@ def check(self):
         return False
 
 
+class TaskStatusCondition(Condition):
+    def __init__(self, agent, task_memid, status="finished"):
+        super().__init__(agent)
+        self.status = status
+        self.task_memid = task_memid
+
+    def check(self):
```
anuragprat1k, line 82: This returns true when the task is finished, running or paused and never returns False (since it will raise an assertion error before that)? Why is that?

base_agent/loco_mc_agent.py
```diff
@@ -218,24 +218,23 @@ def step(self):
...
     def task_step(self, sleep_time=0.25):
-        # Clean finished tasks
-        while (
-            self.memory.task_stack_peek() and self.memory.task_stack_peek().task.check_finished()
-        ):
-            self.memory.task_stack_pop()
-
-        # If nothing to do, wait a moment
-        if self.memory.task_stack_peek() is None:
+        query = {"base_table": "Tasks", "base_range": {"minprio": -0.5, "maxpaused": 0.5}}
```
anuragprat1k, line 221: this looks significantly more involved than before. Please add documentation for this. What do the two queries represent? - ` query = {"base_table": "Tasks", "base_range": {"minprio": -0.5, "maxpaused": 0.5}}` and `query = {"base_table": "Tasks", "base_range": {"minrunning": 0.5}}`

base_agent/task.py
```diff
@@ -28,56 +25,154 @@ class Task(object):
...
+
+    Args:
+        agent: the agent who will perform this task
+        task_data (dict): a dictionary stores all task related data
+    """
+
+    def __init__(self, agent, task_data):
+        super().__init__(agent)
+        for task in task_data.get("new_tasks_fn"):
+            self.child_generator.append(task)
```
anuragprat1k, line 110: what is the notion of children here? Do tasks generate children - how?

anuragprat1k (changes requested):
Thanks for sharing this. I took a look, but do not understand it nearly as deeply as you've written it from the context and documentation provided. Left some questions to try and understand it more. I'm happy to help do tasks here if you can generate them - in fact I'd highly recommend doing that. It will help write and review these large changes in a way that maximizes the understanding of the group.

The major differences from the status quo as I understand are (and please check me on these) - 
1. Tasks are stored in memory through `ControlBlock` in the base interpreter instead of being pushed on the task stack at different places
2. `task_step` of the controller is more involved now - it deals with stepping through the tasks based on priority and state.
3. Tasks can now generate children (`child_generator` in `ControlBlock`)? 
4. Anything else?

base_agent/base_memory_schema.sql
```diff
@@ -93,10 +93,19 @@ CREATE TABLE Programs (
...
+-- prio = 0 :  check my start condition, run if true
+-- prio < 0 :  don't even check my start condition
```
kavyasrinet, line 99: why do we need to keep these ?

base_agent/condition.py
```diff
@@ -22,6 +22,27 @@ def check(self):
         return False
 
 
+class AlwaysCondition(Condition):
```
kavyasrinet, line 25: another reminder that these need to be added to the spec :)

base_agent/condition.py
```diff
@@ -52,6 +73,33 @@ def check(self):
...
 
+class TaskStatusCondition(Condition):
+    def __init__(self, agent, task_memid, status="finished"):
+        super().__init__(agent)
+        self.status = status
+        self.task_memid = task_memid
+
+    def check(self):
+        T = None
+        if self.agent.memory.check_memid_exists(self.task_memid, "Tasks"):
```
kavyasrinet, line 84: what if the task isn't in the memory ? We should return False here

base_agent/dialogue_objects/interpreter.py
```diff
@@ -14,6 +14,31 @@
...
 
 from base_agent.base_util import ErrorWithResponse, NextDialogueStep
+from base_agent.task import ControlBlock
+from base_agent.memory_nodes import TripleNode, TaskNode
+from base_agent.condition import NTimesCondition
+
+
+class TaskListWrapper:
+    """
+    takes a callable that outputs a list (of tasks) and turns it into a callable
```
kavyasrinet, line 24: The docstring should be rephrased to make the functionality clearer. Just looking at the class, I don't quite understand what the purpose of that is is and why we need it. Can we make the documentation clearer ?

base_agent/dialogue_objects/interpreter.py
```diff
@@ -14,6 +14,31 @@
...
+    that outputs one element of the list at a time, iterating through the list
+    """
+
+    def __init__(self, new_tasks_fn):
+        self.list_fn = new_tasks_fn
```
kavyasrinet, line 29: What's the purpose of list_fn ?

base_agent/dialogue_objects/interpreter.py
```diff
@@ -85,17 +113,42 @@ def step(self) -> Tuple[Optional[str], Any]:
...
+                    return err.chat, None
+                tasks_to_push.extend(tasks)
+            if len(tasks_to_push) == 1:
+                TaskMem = TaskNode(self.agent.memory, tasks_to_push[0].memid)
+            elif len(tasks_to_push) > 1:
+                task_data = {"new_tasks_fn": tasks_to_push}
+                c = ControlBlock(self.agent, task_data)
+                TaskMem = TaskNode(self.agent.memory, c.memid)
+            else:
+                TaskMem = None
```
kavyasrinet, line 143: we can get rid of this else if we just init `TaskMem = None` before starting with the check on length

base_agent/dialogue_objects/interpreter.py
```diff
@@ -85,17 +113,42 @@ def step(self) -> Tuple[Optional[str], Any]:
...
+            if TaskMem:
+                TaskMem.get_update_status({"prio": 1})
```
kavyasrinet, line 145: we should document the possible values of `prio` , what they represent and perhaps call it `priority` for better variable naming

base_agent/dialogue_objects/interpreter.py
```diff
@@ -158,18 +211,17 @@ def new_tasks():
                 self.loop_data = mems[0]
             steps, reldir = interpret_relative_direction(self, location_d)
 
+            # FIXME grammar to handle "remove" vs "stop"
+
             loop_task_data = {
-                "new_tasks_fn": new_tasks,
-                "stop_condition": condition,
+                "new_tasks_fn": [new_tasks],
```
kavyasrinet, line 217: reminder that this needs to be revisited once we decide on grammar

base_agent/loco_mc_agent.py
kavyasrinet, line 221: we need a documentation describing what these numbers and keys mean and what they are doing

base_agent/loco_mc_agent.py
```diff
@@ -218,24 +218,23 @@ def step(self):
...
+        task_mems = self.memory.basic_search(query)
+        for mem in task_mems:
+            if mem.task.on_condition.check():
+                # eventually we need to use the multiplex filter to decide what runs
+                mem.get_update_status({"prio": 1, "running": 1})
+            if mem.task.stop_condition.check():
+                mem.get_update_status({"prio": 0, "running": 0})
+        query = {"base_table": "Tasks", "base_range": {"minrunning": 0.5}}
```
kavyasrinet, line 229: same comment as above ^

base_agent/memory_nodes.py
```diff
@@ -763,22 +762,41 @@ class TaskNode(MemoryNode):
...
         super().__init__(agent_memory, memid)
-        pickled, created_at, finished_at, action_name = self.agent_memory._db_read_one(
-            "SELECT pickled, created_at, finished_at, action_name FROM Tasks WHERE uuid=?", memid
-        )
+        try:
+            pickled, prio, running, created_at, finished_at, action_name = self.agent_memory._db_read_one(
+                "SELECT pickled, prio, running, created_at, finished_at, action_name FROM Tasks WHERE uuid=?",
+                memid,
+            )
+        except:
```
kavyasrinet, line 785: we need better error handling here and shouldn't check in our ipdb set_trace :)

base_agent/memory_nodes.py
```diff
@@ -789,17 +807,92 @@ def create(cls, memory, task: Task) -> str:
...
+            if k == "finished":
+                if self.task.finished:
+                    status_out[k] = self.agent_memory.get_time()
+                else:
+                    status_out[k] = -1
+
+            if status.get(k) or force_db_update:
+                key = k
+                if k == "finished":
+                    key = "finished_at"
```
kavyasrinet, line 873: we can just save this info in `status_out ` with the key `finished_at` to avoid confusion here

base_agent/task.py
```diff
@@ -28,56 +25,154 @@ class Task(object):
...
+    def check_remove_and_running_children(stepfn):
+        def modified_step(self):
+            if self.remove_condition.check():
+                self.finished = True
+            if self.finished:
+                TaskNode(self.agent.memory, self.memid).get_update_status(
+                    {"prio": -1, "finished": True}
+                )
+                return
+            query = {
```
kavyasrinet, line 55: same comment as above for this

base_agent/task.py
```diff
@@ -28,56 +25,154 @@ class Task(object):
...
+    """
+
+    def __init__(self, agent, task_data):
+        super().__init__(agent)
+        for task in task_data.get("new_tasks_fn"):
+            self.child_generator.append(task)
+        self.stop_condition = task_data.get("stop_condition", NeverCondition(None))
+        self.on_condition = task_data.get("on_condition", AlwaysCondition(None))
+        self.remove_condition = task_data.get(
+            "remove_condition", TaskStatusCondition(agent, self.memid)
```
kavyasrinet, line 114: reminder that this will need to be updated to be in sync with grammar

kavyasrinet (changes requested):

kavyasrinet:
On a high level - the majority of the changes proposed here and the decisions around them aren't very clear to me. I think it's important to set the context of events and conditionals here for all reviewers.

base_agent/sql_memory.py
aszlam: there are a lot of pieces of Tasks and other memories that are not pickleable for various reasons (e.g. connection to the database); most of these predate this PR or change.  I just grouped them into an explicit const rather than a separate list in each function, like in https://github.com/facebookresearch/droidlet/blob/main/base_agent/sql_memory.py#L1139.

base_agent/memory_nodes.py
aszlam: I can change the name; it _gets_ and _updates_ (so get_update)

base_agent/memory_nodes.py
aszlam: task is stepped in main agent loop; this just says what the statuses mean

base_agent/task.py
aszlam: this a wrapper so we don't have to write the check for children etc every time we write a new task step; it takes in a stepfn (function) and returns a modified_step (function)

base_agent/task.py
aszlam: not sure what you mean by "scope"?  the control block implements the basic event control (starting and stopping child tasks when conditions are met).  will beef up docstrings

base_agent/memory_nodes.py
aszlam: then the input status dict would be like "finished_at": True instead of "finished": True, that's confusing to me?  We could changed the db column to be "finished" instead of "finished_at" more generally

base_agent/memory_nodes.py
aszlam: lol

base_agent/loco_mc_agent.py
aszlam: this is a good example of the SQL++ discussion from a couple of days ago.  This query in SQL would be self documenting; but then would be raw SQL in the agent (and so would break if the backend db was switched to some other key-value store etc.   If we had our own dialect of SQL and parsed it (and made our own interface), all the queries would be clearer...  fwiw its "SELECT * FROM Tasks WHERE prio > -.5 and paused < .5", so select all tasks where the prio allows the on_condition to be checked and where the task has not been explicitly paused by the player

base_agent/dialogue_objects/interpreter.py
```diff
@@ -85,17 +113,42 @@ def step(self) -> Tuple[Optional[str], Any]:
...
+                    if len(r) == 3:
+                        tasks, response, dialogue_data = r
+                    else:
+                        # FIXME don't use this branch, uniformize the signatures
+                        tasks = []
+                        response, dialogue_data = r
+                except ErrorWithResponse as err:
+                    return err.chat, None
+                tasks_to_push.extend(tasks)
+            if len(tasks_to_push) == 1:
```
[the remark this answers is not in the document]
aszlam: if y'all are sure that's what you want?  the majority of tasks we currently push won't have a control block

base_agent/condition.py
aszlam: lol its a bug

base_agent/condition.py
aszlam: see response to Anurag above- if the task isn't in memory and its a "finished" check, the condition should return True.  if it is not in memory and it is a "running" or "paused" check, the branch will do nothing, and the final return will be False

base_agent/memory_nodes.py
kavyasrinet: Sure the latter sounds good.

base_agent/documents/logical_form_specification/events.md
```diff
@@ -1,10 +1,113 @@
...
+	    
+    if remove_condition:
+        finished = True
+```
+
+The "control" block and the "spatial_control" blocks are optional.  Within the control block, each key is optional.
+The defaults (if a key is not set) are "init_condition": ALWAYS, "run_condition": ALWAYS, "stop_condition": NEVER.
+
+The default for "remove_condition" is a bit more involved. Note that conditions are checked at each agent.step(),
+and an "ALWAYS" would then immediately remove the command.  While we can check conditions on the task, the FILTERS
```
kavyasrinet, line 50: here do we mean "ALWAYS" as the value for "remove_condition" ?

base_agent/documents/logical_form_specification/events.md
```diff
@@ -1,10 +1,113 @@
...
+```
+THIS_ACTION_SEQUENCE = {"memory_type": "TASKS",
+                        "triples": [{pred_text:"has_tag", "obj_text": "THIS_SEQUENCE"}]}
+
 ```
+(we will need to update this when we have "obj_text": {"fixed_value": "THIS_SEQUENCE"}).  Furthermore, for each Task in memory, we store the
+number of times it has run; and so have the attribute "RUN_COUNT".  The default (that is, what would be interpreted
+if the key is not set) for remove_condition is 
+```
+THIS_SEQUENCE_FINISHED = {"condition": {
```
kavyasrinet, line 63: do we need the "condition" key here if the parent key is "remove_condition" ?

base_agent/documents/logical_form_specification/events.md
```diff
@@ -1,10 +1,113 @@
...
+                           "comparator": {
+			     "comparison_type": "EQUAL",
+                             "input_left": {
+			       "value_extractor": {
+			         "filter": {
+			           "memory_type": "TASKS",
+				   "triples": [{pred_text:"has_tag", "obj_text": "THIS_SEQUENCE"}],
```
kavyasrinet, line 70: proposed nit "THIS_SEQUENCE" -> "CONTROL_COMMAND_SEQUENCE"

kavyasrinet (commented):
Thanks for the spec update. I left some comments.

base_agent/documents/logical_form_specification/events.md
```diff
@@ -1,10 +1,113 @@
 Note: Classes used in the dict using `<>` are defined as subsections and their corresponding dicts should be substituted in place.
 
 # EVENT draft #
-The "remove" key describes when the even should be deactivated/removed from memory; the "control" clause describes when the event fires.  events are siblings of "action_sequence" and "dialogue_type"
+``` 
+COMMAND = {"action_sequence" : [ACTION/COMMAND, ...., ACTION/COMMAND],
```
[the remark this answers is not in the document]
aszlam: ok

base_agent/documents/logical_form_specification/events.md
```diff
@@ -1,10 +1,113 @@
...
+	    agent_set_action_sequence_priority_to_runme
+	if stop_condition:
+	    agent_set_action_sequence_priority_to_check
+	    
+    if remove_condition:
+        finished = True
+```
+
+The "control" block and the "spatial_control" blocks are optional.  Within the control block, each key is optional.
+The defaults (if a key is not set) are "init_condition": ALWAYS, "run_condition": ALWAYS, "stop_condition": NEVER.
```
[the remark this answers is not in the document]
aszlam: remove_condition would end the control block and command.  the stop_condition just says don't do this.  its a different naming scheme then we have been using: stop just reverts the priority back to 0, whereas remove finishes the command entirely (sets its priority to -1)

base_agent/documents/logical_form_specification/events.md
aszlam: yes, if the remove_condition were "ALWAYS" it would be immediately removed

base_agent/documents/logical_form_specification/events.md
aszlam: nope!  thats a bug

base_agent/memory_attributes.py
```diff
@@ -32,6 +42,62 @@ def __repr__(self):
...
+                            next_step.append(self.agent.memory.get_mem_by_id(n[0][2]))
+                if len(n) == 0:
+                    next_step.append(None)
+        return next_step
+
+    def __repr__(self):
+        return "triple path: " + str(self.path)
+
+
+class AttributeSequence(Attribute):
```
kavyasrinet, line 87: when is this used ?

craftassist/agent/dialogue_objects/mc_interpreter.py
```diff
@@ -380,32 +379,34 @@ def new_tasks():
...
-        print("inside dig, dict", d)
         if "stop_condition" in d:
-            print("stop condition", d["stop_condition"])
             condition = self.subinterpret["condition"](self, speaker, d["stop_condition"])
-            self.append_new_task(
-                self.task_objects["loop"],
-                {"new_tasks_fn": new_tasks, "stop_condition": condition, "action_dict": d},
+            return (
+                [
+                    self.task_objects["control"](
```
kavyasrinet, line 394: this will need to be updated with the spec change since we don't have the "control" key anymore ?

craftassist/agent/dialogue_objects/mc_interpreter.py
```diff
@@ -491,22 +492,29 @@ def new_tasks():
...
+            return maybe_task_list_to_control_block(tasks_to_do, self.agent)
 
         if "stop_condition" in d:
             condition = self.subinterpret["condition"](self, speaker, d["stop_condition"])
-            self.append_new_task(
-                self.task_objects["loop"],
-                {"new_tasks_fn": new_tasks, "stop_condition": condition, "action_dict": d},
+            return (
+                [
+                    self.task_objects["control"](
```
kavyasrinet, line 501: same as above ^

locobot/agent/dialogue_objects/loco_interpreter.py
```diff
@@ -161,26 +160,30 @@ def new_tasks():
...
 
         if "stop_condition" in d:
-            condition = self.subinterpret["stop_condition"](self, speaker, d["stop_condition"])
-            self.append_new_task(
-                self.task_objects["loop"],
-                {"new_tasks_fn": new_tasks, "conditions": condition, "action_dict": d},
+            condition = self.subinterpret["condition"](self, speaker, d["stop_condition"])
+            return (
+                [
+                    self.task_objects["control"](
```
kavyasrinet, line 169: same as above ^

kavyasrinet (approved):
Left few comments but overall this is a huge change. I am okay with merging this sooner rather than later if tests pass - with the option of a potential refactor if folks want. 
You should also create issues and tasks for others on the team so everyone can understand the change better @aszlam .

base_agent/dialogue_objects/filter_helper.py
```diff
@@ -53,6 +53,125 @@ def get_pos(self):
...
+    output = filters_d.get("output")
+    val_map = None
+    if output and type(output) is dict:
+        attr_d = output.get("attribute")
+        get_attribute = interpreter.subinterpret.get("attribute", AttributeInterpreter())
+        a = get_attribute(interpreter, speaker, attr_d, get_all=get_all)
+        val_map = ApplyAttribute(interpreter.agent.memory, a)
+    elif output and output == "count":
+        val_map = CountTransform(interpreter.agent.memory)
+    return val_map
```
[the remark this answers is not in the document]
aszlam: then this will output None, and farther down
https://github.com/facebookresearch/droidlet/blob/812c4f63b7e0804a6aa44ed206de9445459240e9/base_agent/dialogue_objects/filter_helper.py#L206
nothing will be appended to the head of the FILTER, and so it will just output the memory

craftassist/agent/dialogue_objects/mc_interpreter.py
aszlam: this is not part of spec:  its the Task Object that implements control (e.g. a ControlBlock).  see https://github.com/facebookresearch/droidlet/blob/812c4f63b7e0804a6aa44ed206de9445459240e9/craftassist/agent/dialogue_objects/mc_interpreter.py#L93

base_agent/memory_attributes.py
aszlam: here: https://github.com/facebookresearch/droidlet/blob/812c4f63b7e0804a6aa44ed206de9445459240e9/base_agent/dialogue_objects/attribute_helper.py#L164

for example, in order to answer questions about the Task

craftassist/test/test_conditions.py
```diff
@@ -100,19 +94,21 @@ def test_move_till_condition(self):
...
+#        # it is now if everything else works, but should force it
+#        d = deepcopy(
+#            STOP_CONDITION_COMMANDS["follow the cow for 18 seconds after it has x greater than 5"]
+#        )
+#        start_time = self.agent.get_time()
+#        self.handle_logical_form(d, max_steps=5000)
+#        end_time = self.agent.get_time()
+#        time_elapsed = (end_time - start_time) / TICKS_PER_SEC
+#        self.assertEqual(cow_move_sequence[self.agent.world.count - 1 - 18][0][0], 6)
+#        self.assertLessEqual(
```
anuragprat1k, line 109: uncomment if not intentional.

base_agent/documents/logical_form_specification/events.md
```diff
@@ -1,10 +1,115 @@
...
+  }
+}
+
+```
+
+We don't _really_ need 4 conditions; and could probably merge run_condition and stop_conditions, but separating allows each of
+the conditions to be simpler (for example when the condition for running is different from the condition for stopping).
+
+
+TODO modifiers (faster/slower, happily, etc...)
```
anuragprat1k, line 115: this MD is useful, thanks! Reminder to generate feature-level tasks for follow-ups after this.

anuragprat1k (approved):

craftassist/test/test_conditions.py
aszlam: the test is actually flaky.  will make a task to fix it

base_agent/task.py
```diff
@@ -21,63 +56,213 @@ class Task(object):
...
+        TaskNode(agent.memory, self.memid).update_task(task=self)
+        # TODO if this is a command, put a chat_effect triple
 
-    def step(self, agent):
-        """The actual execution of a single step of the task is defined here."""
-        # todo? make it so something stopped by condition can be resumed?
-        if self.stop_condition.check():
-            self.finished = True
+    @staticmethod
+    def step_wrapper(stepfn):
```
soumith, line 93: this implementation has a bug. If `step(...)` takes arguments, for example https://github.com/facebookresearch/droidlet/blob/67bfa83d422540b02f5a3ca729a65921927cf557/locobot/agent/tasks.py#L346 , then it errors out.

I'll send a patch

Issue #145, opened 2021-02-12
Dialogue manager JSON schema validation


Rebecca-Qian:
## Type of Issue

Select the type of issue:
- [ ] Bug report (to report a bug)
- [x] Feature request (to request an additional feature)
- [ ] Tracker (I am just using this as a tracker)
- [ ] Refactor request
- [ ] Documentation Ask

## Description

Check JSON schema for logical forms loaded by dialogue manager.

Issue #117, opened 2021-02-09
Fix module paths in the repo


Rebecca-Qian:
## Type of Issue

Select the type of issue:
- [ ] Bug report (to report a bug)
- [ ] Feature request (to request an additional feature)
- [ ] Tracker (I am just using this as a tracker)
- [x] Refactor request
- [ ] Documentation Ask

## Description

There are still a lot of `sys.path.append`s in the repo. I suggest we permanently add shared modules to `sys.path`. 2 ways to do this:
1.  `site` module, which reads a list of directories and appends them to the module search path
https://docs.python.org/3/library/site.html

2. `PYTHONPATH` environment variable
https://docs.python.org/3/tutorial/modules.html#the-module-search-path

My vote is option #1.

## Current Behavior

`sys.path` is manipulated in Python sessions with `sys.path.append`.

## Expected Behavior

Being able to globally import from any `base_agent` file. Solution can be a single `.pth` file containing directory paths to automatically import to `sys.path`  on initialization.

soumith:
we should completely remove `sys.path.append` everywhere.

We now have a `setup.py`, so our development really should be to execute `python setup.py develop` and then do development.

PR #149, opened 2021-02-17
Validate parse trees in dialogue manager + modules path fix

Rebecca-Qian:
# Description

Add a hook in dialogue manager to validate parse trees from ground truth or semantic parser.

If the parse tree fails the grammar spec, treat as NOOP (log details).

Added a new dialogue manager unit test to test `validate_parse_tree` method. 

PR also contains some fixes for module paths, see #117. 

Fixes # (issue)
#145 
#117 (partial fix)

## Type of change

- [ ] Refactor
- [x] This is a unit test

## Type of requested review

## Before and After

Example of action dictionary caught by validator when playing with bot (later fixed in ground truth):
P176155293

# Testing

In addition to conda setup, make sure modules are properly initialized with
`python setup.py develop`

Then run unit tests and test the bot as usual.

# Checklist:

### DIFF
--- a/base_agent/nsp_dialogue_manager.py
+++ b/base_agent/nsp_dialogue_manager.py
@@ -237,7 +237,7 @@ class DialogModel:
         """
         # RefResolver initialization requires a base schema and URI
         schema_dir = "{}/".format(pkg_resources.resource_filename('base_agent.documents', 'json_schema'))
-        json_validator = JSONValidator(schema_dir, span_type="both")
+        json_validator = JSONValidator(schema_dir, span_type="all")
         is_valid_json = json_validator.validate_instance(parse_tree)
         return is_valid_json
 
--- a/base_agent/nsp_dialogue_manager.py
+++ b/base_agent/nsp_dialogue_manager.py
@@ -33,7 +33,6 @@ from base_util import hash_user
 
 spacy_model = spacy.load("en_core_web_sm")
 
-# import ipdb; ipdb.set_trace()
 class NSPDialogueManager(DialogueManager):
     """Dialogue manager driven by neural network.
 
@@ -231,7 +230,6 @@ class DialogModel:
             from ttad.ttad_transformer_model.query_model import TTADBertModel as Model
 
             self.model = Model(model_dir=ttad_model_dir, data_dir=data_dir)
-            # import ipdb; ipdb.set_trace()
         else:
             raise NotADirectoryError
 
--- a/base_agent/nsp_dialogue_manager.py
+++ b/base_agent/nsp_dialogue_manager.py
@@ -7,6 +7,7 @@ import os
 import re
 import spacy
 import ast
+import pkg_resources
 from typing import Tuple, Dict, Optional
 from glob import glob
 from jsonschema import validate, exceptions, RefResolver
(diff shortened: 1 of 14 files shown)


Rebecca-Qian:
FYI, removed `spacy` from `craftassist/requirements.txt` because it conflicts with base_agent dependencies. See P176017299 for error trace.

Doesn't seem to cause issues for me, cc @snyxan @kavyasrinet

Rebecca-Qian:
Update: also fixed import paths in locobot tests, since some tests were failing.

cc @anuragprat1k

base_agent/nsp_dialogue_manager.py
```diff
@@ -269,6 +294,16 @@ def get_logical_form(self, s: str, chat_as_list=False, ground_truth_actions: dic
             else:
                 d = self.model.parse(chat=s)
 
+        # Validate parse tree against grammar
+        is_valid_json = self.validate_parse_tree(d)
```
anuragprat1k, line 298: how frequently do you expect this to be invalid? do we want to collect all the invalid parses - if so how?

craftassist/test/test_dialogue_manager.py
```diff
@@ -80,6 +80,11 @@ def test_parses(self):
                 )
             )
 
+    def test_validate_json(self):
```
anuragprat1k, line 83: add a successful validation test scenario as well.

craftassist/test/validate_json.py
```diff
@@ -27,7 +27,16 @@ def validate_data(data_path, schema, resolver):
             except exceptions.ValidationError as e:
                 print(command)
                 pprint(parse_tree)
-                print("\n")                            
+                print("\n")
+
+
+def validate_instance(parse_tree, schema, resolver):
```
anuragprat1k, line 33: where is this used?

anuragprat1k (changes requested):
The json validation changes look good. Requested some minor changes.

I don't see any of the `sys.path.append`s removed in this PR, so I don't understand why the tests were failing - did you figure out why?

kavyasrinet (commented):
Thanks for making the change. Looks good. Let's add a unit test too.

base_agent/nsp_dialogue_manager.py
Rebecca-Qian: For ground truth, this should never be invalid, if it is, we need to fix ASAP. For parses from the semantic parser, it should not happen frequently. The model is generally good at producing parse trees that fit the grammar spec, even when they're wrong.

Agree, the wrong parses should be collected. I'll write to a separate file in datasets.

craftassist/test/validate_json.py
Rebecca-Qian: Added docstring here

Rebecca-Qian:
Updates:

- Added 2 test cases checking for spans with string values and arrays
- Added support for string value spans, since these are the format of spans in unit tests and currently autocomplete outputs. I think it's more readable to keep hand written parses in this format, so I decided to support different validation formats rather than enforce array spans everywhere.
- Refactored JSON validator and added docstrings

base_agent/nsp_dialogue_manager.py
Rebecca-Qian: Actually, may not need to write the failed parses to a separate file. Considering adding data checks to CI, then ground truth would always be up to date. Failed model parses aren't very actionable

anuragprat1k (approved):
looks good.

PR #195, opened 2021-02-24
Spec update to update 'repeat's as 'stop_condition'

kavyasrinet:
# Description

This PR proposes the spec update to replace `repeat`s with `stop_condition` when `RUN_COUNT` of the current task is equal to a given loop count.

This PR build on top of spec changes proposed in : 
https://github.com/facebookresearch/droidlet/blob/5396488cd564aad45fb09d4ebb4fedcca3c5952b/base_agent/documents/logical_form_specification/events.md on events. 
We should merge the spec in the above PR first.

Note that we will need to update the datasets to reflect this change.

Fixes # (issue)

## Type of change

- [ ] Refactor
- [ ] This is a unit test

## Type of requested review

## Before and After

Before:
"dance 5 times"
```
{
  "action_type" : "DANCE",
  "repeat" : {
      "repeat_key" : "FOR",
      "repeat_count" : [0, [1, 1]]
    }
}
```
Now:
```
{
  "action_type" : "DANCE",
  "stop_condition" : {
      "condition_type" : "COMPARATOR",
      "condition" : {
          "comparison_type" : "EQUAL",
          "input_left": { "filters" : {
                "output" : {"attribute" : "RUN_COUNT"},
                "memory_type" : "_THIS"
                } } ,
         "input_right" : {"value" : span }
         }
    }
}
```

# Checklist:

### DIFF
--- a/base_agent/documents/logical_form_specification/conditions.md
+++ b/base_agent/documents/logical_form_specification/conditions.md
@@ -12,8 +12,6 @@ Where
 
 ```
 CONTROL = {
-      "repeat_count" : span,
-      "repeat_key" : 'FOR'/'ALL',
       "on_condition": CONDITION, 
       "stop_condition": CONDITION
       }
@@ -58,6 +56,7 @@ The "action_type" key specifies a single action and is mutually exclusive with t
 
 
 ## CONDITION ##
+
 ```
 CONDITION = {
   "condition_type" : 'MEMORY' / 'COMPARATOR' / 'NEVER' / 'TIME' / 'ACTION' /
@@ -98,3 +97,18 @@ TIME_CONDITION = {
 ```
 Note that `event` in time_condition represents when to start the timer and
 `input_left` by default in time condition marks the time since the event condition.
+
+
+Note for refactor: We can perhaps remove the "condition_type" field and have the value of "condition" start with name of condition like for example:
+```
+CONDITION = {
+  "and_condition": [CONDITION, ..., CONDITION]} 
+}
+```
+instead of 
+```
+CONDITION = {
+  "condition_type" : "AND",
+  "condition" : {"and_condition": [CONDITION, ..., CONDITION]} }
+}
+```
--- a/base_agent/documents/logical_form_specification/human_give_command.md
+++ b/base_agent/documents/logical_form_specification/human_give_command.md
@@ -20,11 +20,12 @@ This is the action to Build a schematic at an optional location.
 { "action_type" : BUILD,
   <Location>,
   <Schematic>,
-  <Repeat> (with repeat_key: 'FOR' and additional repeat_dir: 'SURROUND'),
+  <Repeat>,
   "replace" : True
 }
-    
 ```
+Note also that: we can have an additional spatial_control: 'SURROUND' to replace `repeat_dir` when events and conditionals spec change is in.
+
 ### Copy Action ###
 This is the action to copy a block object to an optional location. The copy action is represented as a "Build" with an optional reference_object in the tree.
 
@@ -32,7 +33,7 @@ This is the action to copy a block object to an optional location. The copy acti
 { "action_type" : 'BUILD',
   <Location>,
   <ReferenceObject>,
-  <Repeat> (repeat_key = 'FOR'),
+  <Repeat>,
   "replace" : True
 }
 ```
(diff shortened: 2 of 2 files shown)


aszlam:
1: maybe let's make a new field instead of using "memory_type", like a "special" key at the FILTERS root?
2: should be "remove_condition" not "stop_condition" (because after 10 steps, the task should be removed from consideration

aszlam:
so it would be
{
  "action_type" : "DANCE",
  "remove_condition" : {
      "condition_type" : "COMPARATOR",
      "condition" : {
          "comparison_type" : "EQUAL",
          "input_left": { "filters" : {
                "output" : {"attribute" : "RUN_COUNT"},
                "special" : {"fixed_value": THIS"}
                } } ,
         "input_right" : {"value" : span }
         }
    }
}

aszlam:
actually, another thing: I think we should always enclose the action sequence in its control block (which is how the implementation will work).   So instead of having this control block be a sibling of "action_type", imo it should be at the level of the 
```
{ "dialogue_type": "HUMAN_GIVE_COMMAND",
  "action_sequence" : list(<CommandDict>)
}
```
So imo the preferred format should be:
{ "dialogue_type": "HUMAN_GIVE_COMMAND",
  "action_sequence" : [{"action_type" : "DANCE"}],
  "remove_condition" : {
      "condition_type" : "COMPARATOR",
      "condition" : {
          "comparison_type" : "EQUAL",
          "input_left": { "filters" : {
                "output" : {"attribute" : "RUN_COUNT"},
                "special" : {"fixed_value": THIS"}
                } } ,
         "input_right" : {"value" : span }
         }
    }
}

kavyasrinet:
What's the reasoning behind moving the condition to the root compared to action level ?

aszlam:
> What's the reasoning behind moving the condition to the root compared to action level ?

The model I have in my head is that a control block wraps an action or action sequence.  If there is an an action sequence that needs to be looped or turned into an event, then its necessary that the control block is a sibling of "action_sequence"; and so more generally I would like to keep it uniform

kavyasrinet:
Right but I thought we'd also allow it at action level. So for example :
"build 4 cubes and then come to me" should have the "remove_condition" on build and not on the sequence. Does that sound right ?

aszlam:
> Right but I thought we'd also allow it at action level. So for example :
> "build 4 cubes and then come to me" should have the "remove_condition" on build and not on the sequence. Does that sound right ?

yeah, its allowed at the "action" level (but then the "action" level is being converted to "command" level, and you are allowed to have a sequence there too).  So for example  "go to the house and come back three times, and then jump"
would have the command sequence [[go to the house, come back], [jump]]", and the [go to the house, come back] block would have the non-trivial remove condition

kavyasrinet:
>  So for example "go to the house and come back three times, and then jump"

Right and what I am proposing is define the loop / remove condition at command level instead of only sequence.

aszlam:
> > So for example "go to the house and come back three times, and then jump"
> 
> Right and what I am proposing is to also allow defining the loop / remove condition at command level when it makes sense instead of only the primary sequence.
> So in this case the remove_condition will be on the sequence [go to the house, come back] compared to the one above ("build 4 cubes and then come to me") where the remove condition will be on the command "build a cube"

imo the conditions should live at the highest place they can possibly be; and any conditions should always be attached to a _sequence_ of actions (e.g. a command block).  so if you need a condition in the middle of an action sequence, the subsequence (even if its just one) of actions should be a command block).  This removes a niggling ambiguity...

kavyasrinet:
Okay so then what's the proposal for "dance around 4 times and then come to me" ?

aszlam:
> Okay so then what's the proposal for "dance around 4 times and then come to me" ?
depends if you want "extra keys" or not.  I think with no extra keys it looks like
```
DANCE_SEQUENCE = {
  "action_sequence" : [{"action_type" : "DANCE"}],
  "remove_condition" : {
      "condition_type" : "COMPARATOR",
      "condition" : {
          "comparison_type" : "EQUAL",
          "input_left": { "filters" : {
                "output" : {"attribute" : "RUN_COUNT"},
                "special" : {"fixed_value": THIS"}
                } } ,
         "input_right" : {"value" : span }
         }
    }
}
```
```
COME_HERE={               
 "location": { "reference_object": {"special_reference": {"fixed_value": "SPEAKER"}}},
"action_type": "MOVE",
}
```
```
dance around 4 times and then come to me":
{ "dialogue_type": "HUMAN_GIVE_COMMAND",
  "action_sequence": [DANCE_SEQUENCE, COME_HERE],
}
```

kavyasrinet:
Okay this is a much more involved change than the previous proposal then. We didn't define sequences for single actions. The codemod for this will not be straightforward.

aszlam:
> Okay this is a much more involved change than the previous proposal then. We didn't define sequences for single actions. The codemod for this will not be straightforward.

basically I don't think there should ever be a non-trivial sequence _or_ start/stop/init/remove condition without an action_sequence.  but I don't have a strong opinion about it.... I do have a strong opinion on the control block (start/stop/init/remove conditions) living at the highest level possible in the dict

PR #242, opened 2021-03-11
Shasum fix for mac OS

Rebecca-Qian:
Fixes shasum calculation on mac, but there are still discrepancies in hash calculation on mac and linux.

One I identified was `find -printf` causing issues, but there are more issues with data scripts on mac.

Going forward we could 
- only support Linux for data scripts, or
- use a simpler hash procedure (the previous hash calculation was consistent across different OS platforms)


Rebecca-Qian:
For previous hash calculation that supported mac, see

https://github.com/facebookresearch/droidlet/blob/2c5f950f62932e90b337552a8014d3344ea7a6bc/tools/data_scripts/tar_and_hash_datasets.sh

tools/data_scripts/checksum_fn.sh
```diff
@@ -7,7 +7,7 @@ calculate_sha1sum() {
     echo $'\nCalculating checksum for' ${CHECKSUM_FOLDER}
     echo "Saving to ${SAVE_TO_PATH}"
     find $CHECKSUM_FOLDER -type f ! -name '*checksum*' ! -name '*MD*' \
-       -not -path '*/\.*' -printf "%P\n" | sort -z | sha1sum | tr -d '-' \
```
anuragprat1k, line 10: I added `-printf "%P/n"` so that the checksum was based on the relative folder structure and not absolute. For example, the checksum for the datasets folder should be same irrespective of where the folder lives (in craftassist/agent or locobot/agent or anywhere else). Does this new solution achieve that? What is the issue exactly that you were running into?

anuragprat1k (changes requested):
Requesting more details about the issue being fixed. Added some more context around the checksum change.

Rebecca-Qian:
fixed in #249

PR #266, opened 2021-03-19
Load IP dynamically on Mephisto task creation

Rebecca-Qian:
# Description

Changes in this PR
- Set default port to 80, to allow us to use Cloudflare proxies
- Document how to set up `data.csv` for Turk job creation
- Load dashboard IPs dynamically from `data.csv` variables

Fixes # (issue)

## Type of change

- [ ] Refactor
- [ ] This is a unit test

## Type of requested review

## Before and After

Using Mephisto's heroku architect:
Screen Shot 2021-03-19 at 12 15 12 PM

# Testing

To run Mephisto pipeline locally on devfair:

```
python droidlet_static_html_task/static_test_script.py
```

# Checklist:

### DIFF
--- a/crowdsourcing/servermgr/app.py
+++ b/crowdsourcing/servermgr/app.py
@@ -41,7 +41,7 @@ AWS_ACCESS_KEY_ID = os.environ["AWS_ACCESS_KEY_ID"]
 AWS_SECRET_ACCESS_KEY = os.environ["AWS_SECRET_ACCESS_KEY"]
 AWS_DEFAULT_REGION = os.environ["AWS_DEFAULT_REGION"]
 
-port = int(os.environ.get("PORT", 3000))
+port = int(os.environ.get("PORT", 80))
 
 app = Flask(__name__)
 app.register_blueprint(mephisto_router, url_prefix=r"/")
--- a/crowdsourcing/servermgr/app.py
+++ b/crowdsourcing/servermgr/app.py
@@ -41,7 +41,7 @@ AWS_ACCESS_KEY_ID = os.environ["AWS_ACCESS_KEY_ID"]
 AWS_SECRET_ACCESS_KEY = os.environ["AWS_SECRET_ACCESS_KEY"]
 AWS_DEFAULT_REGION = os.environ["AWS_DEFAULT_REGION"]
 
-port = int(os.environ.get("PORT", 80))
+port = int(os.environ.get("PORT", 3000))
 
 app = Flask(__name__)
 app.register_blueprint(mephisto_router, url_prefix=r"/")
--- a/crowdsourcing/README.md
+++ b/crowdsourcing/README.md
@@ -18,13 +18,27 @@ This opens up a task in your browser on localhost, and allows you to view your t
 ## Running Static HTML Tasks with Heroku
 
 ### Pre-reqs
-You need to have a Heroku account and valid Heroku credentials. To install Heroku CLI on Mac OSX, run `brew tap heroku/brew && brew install heroku`. Then log in with `heroku login`.
+You need to have a Heroku account and valid Heroku credentials. To install Heroku CLI on Mac OSX, run `brew tap heroku/brew && brew install heroku`. Then log in with `heroku login` (pass the `-i` flag for validation without browser).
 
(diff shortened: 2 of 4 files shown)


crowdsourcing/servermgr/app.py
```diff
@@ -41,7 +41,7 @@
 AWS_SECRET_ACCESS_KEY = os.environ["AWS_SECRET_ACCESS_KEY"]
 AWS_DEFAULT_REGION = os.environ["AWS_DEFAULT_REGION"]
 
-port = int(os.environ.get("PORT", 3000))
+port = int(os.environ.get("PORT", 80))
```
snyxan, line 44: We don't need to change this: the container port is set here: https://github.com/facebookresearch/droidlet/commit/97375d012e5daec78316fdf3840a03771eeb2fea

snyxan (commented):
Great! Please see the comments.

crowdsourcing/servermgr/app.py
Rebecca-Qian: This was causing issues for me during testing. I believe the servermgr app automatically redirects to 3000 even though the dashboard is served on port 80

soumith: ummm, the linked commit says `MCDASHBOARD_PORT`, but this line is looking for `PORT`

snyxan: Sorry I was confused by the port 80 -- thought it was the dashboard port. It is actually the WebSocketServer port of our heroku app where it communicates with Mephisto. Why does it has to be 80? Could you share the error you got?

soumith: it has to be 80 for CloudFlare reasons.

@Rebecca-Qian it's not useful to change the default to 3000 because as a regular user (i.e. non-sudo) you cant open ports below 1024.

We'll have to pass in that env variable to the environment to set it to 80.

Rebecca-Qian: I don't think we even need to set the WSS port here. It defaults to 80, so just the IP is enough.

Rebecca-Qian: https://tools.ietf.org/id/draft-ietf-hybi-thewebsocketprotocol-09.html#:~:text=By%20default%20the%20WebSocket%20protocol,tunneled%20over%20TLS%20%5BRFC2818%5D.

Rebecca-Qian: This port change is outside the scope of this PR, and is not necessary for v0. I'll submit a follow up if needed.

snyxan: > it has to be 80 for CloudFlare reasons.
> 
> @Rebecca-Qian it's not useful to change the default from 3000 to 80 because as a regular user (i.e. non-sudo) you cant open ports below 1024.
> 
> We'll have to pass in that env variable to the environment to set it to 80.

For ecs instances which are proxied by Cloudflare: yes, it has to be https port.

@Rebecca-Qian I guess my question is: this port (3000->80) is the WebSocketServer (WSS) port where Mephisto communicates with the heroku app it created(monitor health, retrieve log, etc.). This WSS inside heroku app is not proxied by Cloudflare, no?

The workflow of this part is:
1. Mephisto creates a Heroku app inside which a WebSocketServer is started. Mephisto communicates with this Heroku app using WSS via port 3000.  --> This part is not proxied by Cloudflare.
2. What this Heroku app does is: use AWS python lib boto3 to request and start a bunch of ecs instances. Those ecs have dashboard running in it as a service and expose public IPs to Mephisto. Those public IPs are proxied by Cloudflare so that https are supported. Since Cloudflare has port constraints for proxied server, we set it [here](https://github.com/facebookresearch/droidlet/commit/97375d012e5daec78316fdf3840a03771eeb2fea).

Is my understanding of (1) wrong? Do we also proxy the Mephisto-created heroku app as well so that it's required to change WSS port from 3000 to 80?

Rebecca-Qian: Yes I think this is fine, given that this will always be launched w/ heroku i.e. secure connections.

In any case, this discussion doesn't belong in this PR. My bad.

snyxan (approved):

PR #288, opened 2021-03-30
Move repeats to filters of children

kavyasrinet:
# Description

This PR moves repeats inside actions to be inside `"return_quantity"` -> `"random"`

Fixes #219 

## Type of change

Please check the options that are relevant.

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] Proposes a change (non-breaking change that isn't necessarily a bug)
- [ ] Refactor
- [ ] New feature (non-breaking change that adds a new functionality)
- [ ] Breaking change (fix or feature that would break some existing functionality downstream)
- [ ] This is a unit test
- [ ] Documentation only change

## Type of requested review

- [ ] I want a thorough review of the implementation.
- [x] I want a high level review. 
- [ ] I want a deep design review.

## Before and After
`"make two diamond pickaxes"`
BEFORE...
```
{'action_sequence': [{'action_type': 'BUILD',
                      'repeat': {'repeat_count': [0, [1, 1]],
                                 'repeat_key': 'FOR'},
                      'schematic': {'filters': {'triples': [{'obj_text': [0,
                                                                          [3,
                                                                           3]],
                                                             'pred_text': 'has_name'},
                                                            {'obj_text': [0,
                                                                          [2,
                                                                           2]],
                                                             'pred_text': 'has_block_type'}]},
                                    'text_span': [0, [2, 3]]}}],
 'dialogue_type': 'HUMAN_GIVE_COMMAND'}
```
AFTER...
```
{'action_sequence': [{'action_type': 'BUILD',
                      'schematic': {'filters': {'selector': {'return_quantity': {'random': [0,
                                                                                            [1,
                                                                                             1]]},
                                                             'same': 'ALLOWED'},
                                                'triples': [{'obj_text': [0,
                                                                          [3,
                                                                           3]],
                                                             'pred_text': 'has_name'},
                                                            {'obj_text': [0,
                                                                          [2,
                                                                           2]],
                                                             'pred_text': 'has_block_type'}]},
                                    'text_span': [0, [2, 3]]}}],
 'dialogue_type': 'HUMAN_GIVE_COMMAND'}
```

# Testing

Tested locally.

# Checklist:

- [x] I have performed manual end-to-end testing of the feature in my environment.
- [x] I have added Docstrings and comments to the code.
- [ ] I have made changes to existing documentation where needed.
- [ ] I have added tests that show that the PR is functional.
- [ ] New and existing unit tests pass locally with my changes.
- [ ] I have added relevant collaborators to review the PR before merge.

PR #291, opened 2021-04-02
Read autocomplete pattern matches from JSON Schema

Rebecca-Qian:
# Description

Currently autocomplete matches are populated from a hard-coded set of rules in `<path_to_template_tool>/frontend/src/spec/`.

This PR changes the Autocomplete tool to read directly from JSON schema in `base_agent/documents/json_schema/` used to validate logical forms.

Changeset
- Create symlinks to `base_agent/documents/json_schema/` from `<path_to_template_tool>/frontend/src/spec/`.
- Read schema files on app load and create a combined key value store
- Store schema in state and pass to child `LogicalForm` component via props
- Construct autocomplete subtrees from schema, either by grabbing `properties` or collating all options in `anyOf`/`oneOf` conditions
- Style formatting

Follow Up
- This tool purposefully only reads top level keys in the schema. I need to refactor nested definitions and put them in top level definitions, and use references in nested settings. (I don't do any dereferencing in the tool, so definitions have to be explicit to match.) Schema changes will go in a follow up PR after #288.
- Currently the tool supports writing and saving fragments (see image). A separate PR will store fragments as autocomplete paths, render a list of available fragments and support fragment insertion. 

Fixes # (issue)

## Type of change

- [ ] Refactor
- [ ] This is a unit test

## Type of requested review

# Before and After
Minor changes to autocomplete paths
- everything is lower cased, eg. `"MOVE":  ` before would get you the `MOVE` logical form with pre-filled keys, but now you'd need to use `"move":  `
- After inserting the subtree, the full tree needs to be valid JSON. So you can't start with an empty string, you have to start with a dictionary for the root.

# Testing
Screen Shot 2021-04-01 at 5 49 31 PM

# Checklist:


tools/annotation_tools/template_tool/frontend/src/spec/action_dict_components.schema.json
```diff
@@ -0,0 +1 @@
+../../../../../../base_agent/documents/json_schema/action_dict_components.schema.json
```
kavyasrinet, line 1: Should we use the path from root instead ?

tools/annotation_tools/template_tool/frontend/src/spec/other_dialogue.schema.json
```diff
@@ -0,0 +1 @@
+../../../../../../base_agent/documents/json_schema/other_dialogue.schema.json
```
kavyasrinet, line 1: same as above for all json files

kavyasrinet (approved):

tools/annotation_tools/template_tool/frontend/src/spec/action_dict_components.schema.json
Rebecca-Qian: this is a relative path symlink. It's not pretty the alternative with absolute paths is more likely to introduce path resolution bugs, imo.

PR #294, opened 2021-04-05
Fixes to get locobot agents working on main

soumith:
# Description

Found 4 bugs while trying to get locobot agent working on `main`. Commits that fix all four.

## Type of change

## Type of requested review

# Testing

`locobot_agent.py` works again without error

# Checklist:

### DIFF
--- a/base_agent/sql_memory.py
+++ b/base_agent/sql_memory.py
@@ -1011,7 +1011,10 @@ class AgentMemory:
 
     def get_db_log_idx(self):
         """Return log index for database"""
-        return self._db_log_idx
+        try:
+            return self._db_log_idx
+        except AttributeError:
+            return None
 
     def _write_to_db_log(self, s: str, *args, no_format=False):
         """Write to database log file
--- a/locobot/agent/locobot_agent.py
+++ b/locobot/agent/locobot_agent.py
@@ -200,7 +200,8 @@ if __name__ == "__main__":
     # set up stdout logging
     sh = logging.StreamHandler()
     sh.setFormatter(log_formatter)
-    logging.getLogger().addHandler(sh)
+    logger = logging.getLogger()
+    logger.addHandler(sh)
     logging.info("LOG LEVEL: {}".format(logger.level))
     
     # Check that models and datasets are up to date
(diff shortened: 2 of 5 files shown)


base_agent/loco_mc_agent.py
```diff
@@ -188,9 +188,13 @@ def init_controller(self):
         raise NotImplementedError
 
     def handle_exception(self, e):
-        logging.exception(
-            "Default handler caught exception, db_log_idx={}".format(self.memory.get_db_log_idx())
-        )
+        try:
+            err_string = "Default handler caught exception, db_log_idx={}".format(self.memory.get_db_log_idx())
```
anuragprat1k, line 192: here's the root cause - whenever `db_log_path` is `None`, `_db_log_idx` is not initialized (https://github.com/facebookresearch/droidlet/blob/main/base_agent/sql_memory.py#L84). the [db_log_path](https://github.com/facebookresearch/droidlet/blob/main/locobot/agent/locobot_agent.py#L126) is None for the locobot agent. 

just get rid of this try catch entirely and fix the init in sql_memory.py.

anuragprat1k (changes requested):

base_agent/loco_mc_agent.py
soumith: isn't this the right fix though?

`get_db_log_idx()` should raise an Exception if it doesn't exist.
`sql_memory.py` should only set a `_db_log_idx` if there is a database log. Otherwise, that attribute shouldn't be there.

anuragprat1k: imo a class shouldn't be throwing exceptions for variables it manages internally. In the implementation currently, part of the sql_memory.py's behavior is that it throws an `AttibuteError` for getters in a certain code path (when db log in None). A cleaner behavior I think would be to just return `None` from the getter when that is the case instead of throwing an exception that the caller has to deal with.

soumith: fair enough. Ok I went ahead and fixed it in sql_memory.py

anuragprat1k (approved):

PR #314, opened 2021-04-13
nsp_dialogue_manager cleanup 

### LINKED ISSUE #313: Clean up dialogueManager
## Type of Issue

Select the type of issue:
- [ ] Bug report (to report a bug)
- [ ] Feature request (to request an additional feature)
- [x] Tracker (I am just using this as a tracker)
- [ ] Refactor request
- [ ] Documentation Ask

## Description

Clean up `nsp_dialogue_manager.py`

## Checklist

Use this checklist if you are using this issue as a tracker:

- [x] Extract all dataset reads into one function
- [x] safety -> make a set anyway
- [ ] greetings.json -> txt file
- [x] remove ast.literal_eval -> all json.loads
- [x] remove glob
- [x] remove ast
- [x] more documentation to every function and class
- [x] move out gt query from get_logical_form of model
- [ ] remove returning lists from preprocess or loop over the list ? How do we feel about this ? "make 5x5 cubes there. Come here" into :
['make 5 x 5 cubes there', 'come here']
which is two different commands in that sequence.

But the bigger problem is : we aren't even putting that to use since we are just extracting the first chat here: https://github.com/facebookresearch/droidlet/blob/main/base_agent/nsp_dialogue_manager.py#L164

And so technically this is a waste: https://github.com/facebookresearch/droidlet/blob/dbed63f1e6539694f77771e4dd4ce026389ac98c/base_agent/nsp_dialogue_manager.py#L334

If anything, we should loop over the list of chats and process them one by one at the least and if not, we should drop the sentence level tokenizer, but in order to support that we need to allow a list as return value for `maybe_dialog_objs`
- [ ] How should we deal with lists ? Should we handle in model vs NSP_DM ?
- [ ] If model, how do we deal with list in ground truth, safety, greetings
- [ ] Put span resolution and coref resolve as separate methods
- [ ] call them both in handle_logical_form
- [ ] do span resolution and coref resolve for all dashboard returns before returning

kavyasrinet:
# Description

This PR proposes the first attempt at cleaning up `nsp_dialogue_manager.py` in the direction of making it more modular.

Fixes #313 

## Type of change

- [x] Refactor
- [ ] This is a unit test

## Type of requested review

# Changeset

On a high level, this PR:
1. Refactors out dataset reading methods from `init`
2. Moves out `DialogModel` from `NSPDialogueManager`
2. Moves out `NSPLogger` from `NSPDialogueManager`
3. Moves json validation from model into dialogue manager
4. Moves out ground truth check from model's get_logical_form method to dialogue manager's get_logical_form
5. Proposes a method to perform postprocessing of logical form as a separate step outside of model's `get_logical_form`
6. Gets rid of `ast` so we use `json.loads` consistently throughout the codebase
7.  Refactors dialogue_manager.py
8. Introduces a model wrapper

Major refactor described here: https://docs.google.com/document/d/1c42omiUT3Wt7OmWirJHRh-oDBX9PQBTCdwYD76UUkwo/edit 

Please note: `get_last_m_chats` hasn't been implemented yet and is WIP.
Tests are all fixed up.

Remaining refactor:
1. As a followup : `greetings.json` needs to be converted to a txt file so we can keep our datasets in a similar format.

# Testing

Tested locally

# Checklist:


base_agent/dialogue_manager.py
```diff
@@ -38,30 +36,26 @@ class DialogueManager(object):
...
+        chat_list_text = [chat.chat_text for chat in all_chats]
 
-    def is_safe(self, chat):
-        """Check that chat does not contain unsafe words.
-        """
-        safety_set = set(self.safety_words)
-        cmd_set = set(chat.lower().split())
-        notsafe = len(cmd_set & safety_set) > 0
-        return not notsafe
+        return chat_list_text
```
aszlam, line 58: should return (chat_text, speaker) tuples, no?

base_agent/dialogue_manager.py
```diff
@@ -38,30 +36,26 @@ class DialogueManager(object):
         agent: a droidlet agent
         model: a (perhaps ML) model used by and the model used for manager.
     """
-    def __init__(self, agent, model):
+
+    def __init__(
+        self, agent, dialogue_object_classes, opts, semantic_parsing_model_wrapper, model=None
```
aszlam, line 41: what does the model do?  I would have expected only model_wrapper

kavyasrinet: This `model` was from before so I preserved it. from documentation: a (perhaps ML) model used by and the model used for manager.

base_agent/dialogue_manager.py
kavyasrinet: So we already get the speaker info with the chat: https://github.com/facebookresearch/droidlet/pull/314/files/4ff216cb3cafdb5053fd2caa5443ae1054d50d95#diff-79904ff197524a1a7d7aab8130ac6227e88fea0ddef7148377e9cb7503f8ef30R77 

and pass it along to get_dialogue_object() of the model wrapper.

This function only fetches the last m chats (plain text)

aszlam: but that is only the speaker for the incoming chat, no?  presumably if someone is using the multiturn chat history in their semantic parser, they will need the speaker for each of the chats in the history.

kavyasrinet: Bu the speaker is going to be the same in the session(regardless of chat history)right ?

base_agent/dialogue_manager.py
aszlam: ok, so lets torch this one then, if someone wants to update the step with a model they can subclass

kavyasrinet: Sounds good! I can do that. Wanted to keep it around for you to take a look at.

base_agent/dialogue_manager.py
kavyasrinet: Nevermind, I can extract the speaker from here. Let me push a commit

soumith (approved):

PR #333, opened 2021-04-19
Adding 3D bounding boxes for locobot detections

anuragprat1k:
# Description

* Saves 3d bboxes as `(minx, miny, minz, maxx, maxy, maxz)` of the axis aligned bounding box formed from the point cloud for each detection.
* Added these fields to the `DetectedObjectFeatures` table

Fixes # (issue)

## Type of change

- [ ] Refactor
- [ ] This is a unit test

## Type of requested review

# Checklist:


locobot/agent/perception/handlers/core.py
```diff
@@ -72,17 +79,24 @@ class RGBDepth:
...
-        self.gray = cv2.cvtColor(rgb, cv2.COLOR_BGR2GRAY)
 
     def get_pillow_image(self):
         return Image.fromarray(self.rgb, "RGB")
 
+    def get_bounds_for_mask(self, mask):
+        """for all points in the mask, returns the bounds as an axis-aligned bounding box.
+        """
+        indices = zip(*np.where(mask == True))
+        points = [self.get_coords_for_point((x,y)) for x,y in indices]
```
soumith, line 95: these two lines here are obviously super inefficient. do something about it or it'll be a perf bottleneck.

soumith (changes requested):

locobot/agent/perception/handlers/core.py
anuragprat1k: vectorized this, roughly 35x faster now (python snippet below)

locobot/agent/perception/handlers/core.py
```diff
@@ -73,26 +73,28 @@ class RGBDepth:
...
     rgb: np.array
     depth: np.array
-    ptcloud: list
+    ptcloud: np.array
 
     def __init__(self, rgb, depth, pts):
         self.rgb = rgb
         self.depth = depth
-        self.ptcloud = pts
+        self.ptcloud = pts.reshape(rgb.shape)
```
soumith, line 86: are the points in the point cloud laid out in the same co-ordinates as the RGB image? i.e. you should confirm visually that `pts.reshape(rgb.shape)` does what you expect

soumith (approved):

aszlam (approved):

locobot/agent/perception/handlers/core.py
anuragprat1k: Yes. `pts` previously was a (w*h, 3) shaped array (where h and w are height and width of the image respectively) created in row-major form. I wanted to convert it to a (w,h,3) shaped array (which is the shape of the rgb image), so that we have 3 channels (x,y,z) for each pixel in the image. 

Since pts originally is saved in row-major form (or C-like indexing where the last index changes fastest), reshaping works since it uses C-like indexing to reshape the array by default (more [here](https://numpy.org/doc/stable/reference/generated/numpy.reshape.html)). 

Tested this with the following code snippet, and some sanity checking in habitat to gather enough confidence to merge. But it also needs testing on the robot, which I'll do later this week.

Screen Shot 2021-05-05 at 12 18 32 PM

PR #355, opened 2021-05-04
Removes all sys.path hacks, deletes geoscorer

### LINKED ISSUE #117: Fix module paths in the repo

soumith:
# Description

This gets rid off all PYTHONPATH overrides in code and creates a clean package convention.

I converted some imports to absolute and some to relative. A clean up pass can be made later to make this consistent.
Fixes #117 

### Type of requested review: High-level (relying on unit tests)

# Testing

Ran locally:
- unit tests: locobot, craftassist, base_agent
- locobot_agent and craftassist_agent
- examples/grid


kavyasrinet (approved):
LGTM! Thanks so much for taking a stab at this.
We can rely on the tests to be the indicator and if we find anything broken, we can fix on the fly.

Rebecca-Qian:
Thank you, thank you :)

PR #311, opened 2021-04-12
[Hack session] Create protected s3 endpoint for secured assets

Rebecca-Qian:
# Description

Currently, we don't have an internal s3 bucket, i.e. fully blocked from public access. Since configuring buckets to public read causes all objects in that bucket to be publicly accessible, we can't place protected resources in the `craftassist` bucket.

I created `s3://droidlet-internal/` to store secured resources, and moved the command blacklist `safety.txt` to `droidlet-internal`. The protected endpoint will be accessed by droidlet's core team/collaborators, and production systems running agents eg. ECS containers used to run interactive sessions.  

Users are authenticated in 2 ways:
1. `awscli`: intended for droidlet team/collaborators with write access, who have aws CLI configured.
2. environment access keys: intended for remote containers running agent.

Specifics:
- `try_download.sh` calls a script that fetches internal resources for authenticated users
- dialogue manager reads `safety.txt` from `base_agent/documents/internal/` (under .gitignore)
- add `boto3` to root `requirements.txt`
- update unit tests

This PR also updates the datasets hash with a newer release that does not include `safety.txt`. 

## Limitations
- Follow up PR will update NSP datasets with this file removed from ground truth.
- `fetch_internal_resources.py` is kind of scrappy and is supposed to be a short script `try_download.sh` can call. I was tempted to convert all our data scripts to python, but that's way out of the scope of this PR. 

Fixes # (issue)
#243 

## Type of change

Please check the options that are relevant.

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] Proposes a change (non-breaking change that isn't necessarily a bug)
- [ ] Refactor
- [x] New feature (non-breaking change that adds a new functionality)
- [ ] Breaking change (fix or feature that would break some existing functionality downstream)
- [ ] This is a unit test
- [ ] Documentation only change

## Type of requested review

- [x] I want a thorough review of the implementation.
- [ ] I want a high level review. 
- [ ] I want a deep design review.

## Before and After
# Testing

Tested bot end to end, updated safety check in unit tests and all tests pass locally.

Note the "authenticated user" check:

Screen Shot 2021-05-07 at 7 15 33 PM

 
# Checklist:

- [x] I have performed manual end-to-end testing of the feature in my environment.
- [x] I have added Docstrings and comments to the code.
- [x] I have made changes to existing documentation where needed.
- [x] I have added tests that show that the PR is functional.
- [x] New and existing unit tests pass locally with my changes.
- [x] I have added relevant collaborators to review the PR before merge.

Issue #243, opened 2021-03-11
Set up S3 permissions scope for internal only resources


Rebecca-Qian:
## Type of Issue

Select the type of issue:
- [ ] Bug report (to report a bug)
- [ ] Feature request (to request an additional feature)
- [x] Tracker (I am just using this as a tracker)
- [ ] Refactor request
- [ ] Documentation Ask

## Description

One way to do this is to tag objects as internal and give IAM users permissions to access. See

https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-tagging.html

## Current Behavior

craftassist/pubr bucket is set to ACL public read.

## Checklist

- [ ] Create internal scope
- [ ] Authorize relevant IAM users
- [ ] Categorize documents
- [ ] Look into tagging models/dataset versions as `latest` and updating fetch/deploy scripts

Rebecca-Qian:
cc @aszlam , @kavyasrinet

PR #403, opened 2021-05-12
[Bugfix] Fix safety checks failure by adding base_agent/documents/internal/ to version control

Rebecca-Qian:
# Description

Fixes safety test failures caused by adding auth to downloading protected resources. 

The cause for failure is that `safety.txt` was not downloaded to the right place, because `base_agent/documents/internal/` was not tracked as a directory. Git does not track empty directories, and `base_agent/documents/internal/*` path was under our `.gitignore`. To fix, I added a `.gitkeep` and the internal folder to version control.

Fixes # (issue)

#243 #311 (note CI was passing due to bug introduced in previous commits) 

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,6 +33,11 @@ jobs:
             echo ${AWS_S3_ACCESS_KEY_ID} > tools/docker/AWS_S3_ACCESS_KEY_ID.txt
             echo ${AWS_S3_SECRET_ACCESS_KEY} > tools/docker/AWS_S3_SECRET_ACCESS_KEY.txt
 
+      - run:
+          name: Python style checks (black and awscli)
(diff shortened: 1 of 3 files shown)


soumith (approved):
would it make sense instead to have the download script create the directory? I'll let you make the call, approving this one.

kavyasrinet (approved):

Rebecca-Qian:
> would it make sense instead to have the download script create the directory? I'll let you make the call, approving this one.

That works too, and would be a more general solution. I'm using `pkg_resources` to load documents though, so it wouldn't have much advantage here.

PR #424, opened 2021-05-19
Fixing CCache, compatibility with $CC, $CXX, improving speed of circleci builds, improve gflags/glog detection in cmake

soumith:
# Description

There are a few commits, all related to CI / build system.

#### Feature set 1
improve craftassist makefiles by adding $CC / $CXX compatibility, and also make glog/gflags detection more robust

#### Feature set 2
    move minecraft testing base from a circleci docker instance to machine image instance.

    This was done because a circleci docker inside a docker (how the tests were running) has a few limitations, including

    1. only exposing 2 CPUs in the 2nd docker regardless of the actual range of available CPUs
    2. only exposing 8GB of memory regardless of the parent resource class
    3. kernel restrictions that didn't allow one to run any stack sampling tools (such as `perf top` or `py-spy`) in the circleci job via ssh

#### Feature  set 3

    Gets ccache working, fixing 4 different bugs:

    1. this kinda never worked curl http://craftassist.s3-us-west-2.amazonaws.com/pubr/client_ccache.tar.gz because this command aws s3 cp client_ccache.tar.gz s3://craftassist/pubr/client_ccache.tar.gz does not actually make the file public. I had to add the --acl public-read to that command
    2. this RUN export CCACHE_DIR="/cache/.ccache" does not work in the sense that export in a Dockerfile is only alive for the lifetime of the command. So, when the next line in the Dockerfile runs, there is no CCACHE_DIR in the environment variables
    3. the ccache compiler paths are never set, doing which gcc or which g++ still returns /usr/bin/gcc and /usr/bin/g++, so basically ccache is never even entering the picture
    4. The tar extraction of the downloaded ccache was going into `/cache/cache/.ccache` because the `tar cvfz` missed a directory switching `-C` option.

Additionally, upgrades the base craftassist image to Ubuntu 20.04 for improved tooling (for example upgraded `lld` package)

#### Feature  set 4

Removes `mcdeploy.key` in the docker images, which is no longer needed as droidlet is now a public repository

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,13 +22,6 @@ jobs:
             echo 'export DOCKER_BUILD_ARGS="--build-arg current_branch=${CIRCLE_BRANCH} --build-arg AWS_ACCESS_KEY_ID=${AWS_S3_ACCESS_KEY_ID} --build-arg AWS_SECRET_ACCESS_KEY=${AWS_S3_SECRET_ACCESS_KEY}"' >> $BASH_ENV
 
       - run:
-          name: Deploy Github key
-          command: |
-            echo ${MCDEPLOY_KEY_B64} | base64 -d > tools/docker/mcdeploy.key
-            chmod 400 tools/docker/mcdeploy.key
-
-      - setup_remote_docker
-      - run:
           name: Build docker containers
           command: make -C tools/docker build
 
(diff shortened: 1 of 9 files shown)


snyxan (approved):
LGTM. Thanks for fixing the issues and speeding up the build process!

PR #438, opened 2021-05-26
Fix submodule version to stable, working ones as in main branch

snyxan:
# Description

The commits of submodules in this branch are different from those on main, causing some compatibility issues such as cuberite binary path change. Here fix the commits to be the same as those on main branch.

It also seems to solve the issue with shutil copy symlink issue in this commit: https://github.com/facebookresearch/droidlet/commit/cf06fc7537441849442f1d83b4dadfc7f8f1ef5c. @kavyasrinet if I revert this commit and use the updated submodule commits, it will work.

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

Before:

cuberite_process.py can not be run -- failed to find Cuberite executable .

After:

cuberite_process.py can be run successfully

# Testing

Tested on devfair.

**Steps to test**

Since we are updating submodule versions, you should do this before test:

```
git submodule deinit -f .
git submodule update --init --recursive
```

Once you see no chanegs need to be staged using `git status`, run the following commands under root:
```
cd droidlet/lowlevel/minecraft
make cuberite
python cuberite_process.py
```

# Checklist:


kavyasrinet:
How has this been tested ? What are the steps one would need to take once this is merged into `refactor_lowlevel` ?

snyxan:
> How has this been tested ? What are the steps one would need to take once this is merged into `refactor_lowlevel` ?

updated in the Test section.

kavyasrinet:
Wouldn't we need to pull the submodules again since the version has now been updated ?

snyxan:
> Wouldn't we need to pull the submodules again since the version has now been updated ?

Yes, we need to pull the submodules to make sure they' have the correct version. Thanks for the reminder, I've added these part to the test plan.

> Yes confirming that just the test plan doesn't work.
> @snyxan can you please update the test plan with every step that needs to be executed to repro this ?

Do you mean the test plan doesn't work because submodule versions are not updated, or do you mean it doesn't work with a clean state of git (with no changes to be added to stage area )?

The reason I didn't include the submodule update steps at first is because I thought the test plan should be the steps we do with a clean git state (not modified files, not mismatched submodule versions, etc.). But you are right -- since it's a in-place change of submodule versions, I should add it to the test plan so that reviewers can easily repro and test.

kavyasrinet:
Yeah it wasn't working because the PR was being tested against in-place change of submodules. Thanks for updating the test plan. Confirming that I can repro. 
Merging this.

PR #444, opened 2021-06-01
Added unittest for cuberite agent connection

snyxan:
# Description

Added a unittest to test connection between agent and cuberite. It will: 1.start a cuberite subprocess, 2. start an agent process which will try to connect to the cuberite process, 3. check cuberite log to see if agent has successfully connected to it.

Follow-ups:
- More agent actions in the world and test if cuberite gets that (would be tricker with cuberite as a subprocess, but can start with a naive way by checking if certain message is in cuberite log?). wdyt? Like to what extent should we test cuberite?(is connected successfully enough?)

## Type of change

- [ ] Refactor
- [x] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

**Before:**

No unittest to test agent cuberite connection

**After:**

This one tests whether the complied cuberite binary works with agent process.

# Testing

Tested locally with:

```
pytest droidlet/lowlevel/minecraft/
```

# Checklist:


droidlet/lowlevel/minecraft/tests/test_cuberite_agent_connection.py
```diff
@@ -0,0 +1,37 @@
+"""
+Copyright (c) Facebook, Inc. and its affiliates.
+"""
+import os
+import signal
+import subprocess
+import time
+import unittest
+from agents.craftassist.craftassist_agent import CraftAssistAgent
```
kavyasrinet, line 9: I think this test should go under `agents/craftassist/tests/` instead of `lowlevel` because this tests only our current agent with cuberite. Doing so will also preserve what's proposed [here](https://github.com/facebookresearch/droidlet/issues/337) so we don't import `agent` in any of the subfolders going forward.

droidlet/lowlevel/minecraft/tests/test_cuberite_agent_connection.py
```diff
@@ -0,0 +1,37 @@
...
+from agents.craftassist.tests.fake_agent import MockOpt
+
+
+class MyTestCase(unittest.TestCase):
+    def test_agent_cuberite_connection(self):
+        opts = MockOpt()
+        opts.port = 25565
+        cuberite_process_path = "droidlet/lowlevel/minecraft/cuberite_process.py"
```
kavyasrinet, line 17: Nit : once we do the above lets use the standard `os.path.join` instead of hardcoding. So something like :
```
repo_home = os.path.dirname(os.path.realpath(__file__))
cuberite_process_path = os.path.join(repo_home, "../../droidlet/lowlevel/minecraft/cuberite_process.py")
```

kavyasrinet (changes requested):

droidlet/lowlevel/minecraft/tests/test_cuberite_agent_connection.py
snyxan: done

kavyasrinet (approved):

PR #453, opened 2021-06-02
fix attribute check in nsp_model_wrapper

soumith:
# Description

Fixes https://github.com/facebookresearch/droidlet/issues/451

Test incoming...


kavyasrinet:
Thanks for the fix but I tested this with the locobot agent (there are tests for running through locobot with the dialoguemanager and they pass) and it doesn't seem to break. Could this be a python version thing ? In general if an attribute doesn't exist in a Namespace it returns `None` by default.

soumith:
I could reproduce the original issue with Python 3.8.5

anuragprat1k (approved):

kavyasrinet:
Thanks for taking a look!

Issue #101, opened 2021-02-04
[dashboard] teleop commands are inconsistently executed


anuragprat1k:
## Type of Issue

Noticing that the 1 teleop command on the locobot agent can get executed 3-6 times. Is this some quirk of the dashboard [key handler](https://github.com/facebookresearch/droidlet/blob/main/dldashboard/web/src/StateManager.js#L166)? I tried a quick hack of setting the commands array to empty [ after they are emitted ](https://github.com/facebookresearch/droidlet/blob/main/dldashboard/web/src/StateManager.js#L207) but that didn't fix this. 

Select the type of issue:
- [x] Bug report (to report a bug)
- [ ] Feature request (to request an additional feature)
- [ ] Tracker (I am just using this as a tracker)
- [ ] Refactor request
- [ ] Documentation Ask

## Expected Behavior

Each teleop command only gets executed once. 

## Steps to reproduce

Play with the teleop commands on the locobot agent. I encountered this while playing with physical locobot.

soumith:
this is likely a function of how a key-press is registered.

When you press a given key, you are holding it down for a given set of milliseconds in that brief encounter.

This gets registered as multiple key-presses of the same kind.

Usually keyboards / OSes have a software delay before they register a second key.
For example, hold down the key "a". There's a delay between the first "a" and the second "aa" appearing, and after the delay the letters appear at a costant rate.

We can add a software delay to recognize a key-down hold.

kavyasrinet:
Is this still an issue we want to fix or can we close for now ?

anuragprat1k:
We haven't fixed it yet.

pringshia:
@soumith is correct - you may want to explore listening on keyup events instead of keydown events here herehttps://github.com/facebookresearch/droidlet/blob/dbed63f1e6539694f77771e4dd4ce026389ac98c/dldashboard/web/src/components/Navigator.js#L31

if you do want a press-and-hold sort of behavior, you could look into debouncing or throttling the callback function based on the UX you're looking for

hollisma:
After playing around with the movement a bit, it seems like the current implementation moves very similarly to a game, albeit much more laggy and with less immediate feedback, but that's a whole different issue. Even though multiple commands are usually entered with each keypress, I think this is actually fine. If we still want to resolve this issue, I would love feedback on the options I came up with below. Otherwise, I can focus on fine-tuning the current movements (option 3, my preference). 

[...]

PR #458, opened 2021-06-03
fixed duplicate commands from keypresses

hollisma:
# Description

There was an issue where keypresses were resulting in multiple commands being send to the agent. For example, pressing the up arrow key could result in 3+ MOVE_FORWARD commands being processed. The fix implemented in this PR involves processing keypresses on keyups rather than rapidly checking keydowns. 

Fixes issue #101 

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

Before, a keypress would result in multiple commands being processed by the agent. 
After, a keypress will always result in a single command being processed. 

# Testing

All 8 keys were tested and have been verified to send only one command for each keypress, no matter how long the keypress is. For the future, if we want to be able to input multiple commands by holding down a key (like in games), we would need to process both keydowns and keyups. However, the models currently take a long time to process each command, making correctness more important than convenience. 

# Checklist:

### DIFF
--- a/droidlet/dashboard/web/src/StateManager.js
+++ b/droidlet/dashboard/web/src/StateManager.js
@@ -187,9 +187,11 @@ class StateManager {
 
   keyHandler(key_codes) {
     let commands = [];
+    let keys = [];
     for (var k in key_codes) {
       let val = key_codes[k];
       k = parseInt(k);
+      keys.push(k);
       if (val === true) {
         if (k === 38) {
           // Up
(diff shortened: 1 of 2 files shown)


droidlet/dashboard/web/src/StateManager.js
```diff
@@ -227,6 +229,11 @@ class StateManager {
     }
     if (commands.length > 0) {
       this.socket.emit("command", commands);
+
+      // Reset keys to prevent duplicates
```
anuragprat1k, line 233: Why is this needed?

anuragprat1k (approved):
Looks good!

PR #461, opened 2021-06-04
[refactor] clean up perception/robot

soumith:
context and walkthrough provided over video


agents/locobot/locobot_agent.py
```diff
@@ -136,7 +146,33 @@ def init_perception(self):
...
+        x, y, yaw = xyz
+        sio.emit("map", {
+            "x": x,
+            "y": y,
+            "yaw": yaw,
+            "map": self.mover.get_obstacles_in_canonical_coords()
+        })
+
+        previous_objects = DetectedObjectNode.get_all(self.memory)
+        new_state = self.perception_modules["vision"].perceive(rgb_depth,
```
anuragprat1k, line 166: replace this with 
```
new_objects, updated_objects = = self.perception_modules["vision"].perceive(...
```

That way you can also get rid of `if new_state is not None:` statement below.

droidlet/dashboard/web/src/StateManager.js
```diff
@@ -67,7 +66,13 @@ class StateManager {
     this.memory = this.initialMemoryState;
     this.processRGB = this.processRGB.bind(this);
     this.processDepth = this.processDepth.bind(this);
+    this.processRGBDepth = this.processRGBDepth.bind(this);
```
anuragprat1k, line 69: why do we need all of processRGBDepth, processDepth and processRGB?

droidlet/parallel.py
```diff
@@ -0,0 +1,117 @@
...
+def _runner(_init_fn, init_args, _process_fn, shutdown_event, input_queue, output_queue):
+    init_fn = cloudpickle.loads(_init_fn)
+    process_fn = cloudpickle.loads(_process_fn)
+    initial_state = init_fn(*init_args)
+
+    while not shutdown_event.is_set():
+        try:
+            process_args = input_queue.get(block=True, timeout=0.1)
+            process_args_aug = (initial_state, *process_args)
+            print(process_args)
```
anuragprat1k, line 48: replace `print`s with logging

droidlet/parallel.py
```diff
@@ -0,0 +1,117 @@
...
+            self._child_conn.send((e, tb))
+            raise e
+
+    @property
+    def exception(self):
+        if self._parent_conn.poll():
+            self._exception = self._parent_conn.recv()
+        return self._exception
+
+def _runner(_init_fn, init_args, _process_fn, shutdown_event, input_queue, output_queue):
```
anuragprat1k, line 39: why is `_runner` defined here instead of being a private fn of `BackgroundTask`?

droidlet/perception/robot/__init__.py
```diff
@@ -6,37 +6,32 @@
...
     draw_xyz,
     get_random_color,
     get_coords,
     get_color_tag,
-    InputHandler,
-    DetectionHandler,
-    MemoryHandler,
-    FaceRecognitionHandler,
-    ObjectDeduplicationHandler,
+    ObjectDetection,
```
anuragprat1k, line 33: the downside of this naming convention (sans 'handler') is that it is hard to pick out which are structs and which are models (for eg between ObjectDetection vs Detection, HumanPose vs HumanKeypointsOrdering). I prefer more explicit names that convey the intent of the object. One way to fix that would be naming the model-related classes as ObjectDetector, FaceRecognizer and so on. What do you think?

droidlet/tests/test_parallel.py
```diff
@@ -0,0 +1,29 @@
...
+            f = f[0]
+            x = f(b) + 5
+            print(x)
+            return x, x
+    
+        b = BackgroundTask(init_fn=init_fn, init_args=(2,), process_fn=process_fn)
+        b.start()
+        self.b = b
+
+    def forward(self):
```
anuragprat1k, line 22: add this as a unit test too

anuragprat1k (changes requested):
Looks good overall, requesting some minor changes.

droidlet/parallel.py
soumith: the `print` was removed in a later commit. it was kept in this commit by mistake.

droidlet/parallel.py
soumith: because it needs to be picklable. A class-member is not picklable, but a global function is.

droidlet/perception/robot/__init__.py
soumith: We should distinguish them in namespaces rather than via the names of the Structs or Classes themselves. The structs should probably go into a `struct` file then.

droidlet/tests/test_parallel.py
soumith: yes, will do

agents/locobot/locobot_agent.py
soumith: I can't do that I think, because we return `None` when there are no ready `detections`.

See for example:

```
In [1]: def foo(x):
   ...:     if x == 1:
   ...:         return None
   ...:     else:
   ...:         return 1, 2
   ...:

In [2]: a, b = foo(1)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-2f6254e10c77> in <module>
----> 1 a, b = foo(1)

TypeError: cannot unpack non-iterable NoneType object

In [3]: a, b = foo(2)
```

droidlet/dashboard/web/src/StateManager.js
soumith: we can get rid of `processRGBDepth`, I just thought it would be a natural thing for an RGBD object to be logged.

If you look at the implementations, `processRGBDepth` just calls `processRGB` and `processDepth`

droidlet/parallel.py
soumith: on that note, for `init_fn` and `process_fn`, I am using `cloudpickle` precisely because I want to be able to pass class-local or function-local functions into `BackgroundTask` and they are only picklable either via `cloudpickle` or via `dill`, and not with the out of box pickler.

agents/locobot/locobot_agent.py
anuragprat1k: Returning `None, None` would fix that? Don't have a strong opinion, just was thinking of cutting down on lines.

anuragprat1k (approved):

PR #472, opened 2021-06-08
Revert "showing object masks on dashboard"

soumith:
Reverts facebookresearch/droidlet#464


soumith:
cc: @hollisma you probably need to handle `mask=None` case, or the case when mask doesn't exist.

PR #482, opened 2021-06-10
sql syntactic sugar conversion draft

aszlam:
# Description

This is a draft of the new FILTERS and basic_search interface.  


aszlam:
Once y'all look this over and decide it makes sense, I will rewrite basic_filters to use the new syntax (the "sqly" one)

Rebecca-Qian (commented):
This is exciting 😄 

From our offline conversation (cc @kavyasrinet):
- Parser LF linearization expects a tree input, so you can't just give it a SQL string for the filters. I suggest we convert SQL statements into parse trees too, and the vocabulary will be the union of SQL attributes and current tokens.
- Using "SQL" everywhere is misleading because it's not actually SQL. Maybe something like "declarative filters language" or "embodied query language"? Then we can rename "sqly" too.

Also I'm assuming this code will eventually get cleaned up and documented more.

droidlet/memory/filters_conversions.py
```diff
@@ -0,0 +1,549 @@
...
+        "input_right": {"value_extractor": right_value},
+        "comparison_type": ct,
+    }
+
+    return f
+
+
+def maybe_eval_literal(clause):
+    try:
+        output = ast.literal_eval(clause)
```
kavyasrinet, line 394: let's use `json` instead of ast.

kavyasrinet (commented):
Overall, I like the direction of this but can you confirm that the memory APIs (for example basic search) will take this new form as input ?
Few things that are missing:
1. Documentation of the conversion. I think we need function level input -> output mapping documentation and in addition a spec doc that states current filters -> new filters 
2. Unit tests for all the functions when this is in

kavyasrinet:
whoops just saw your comment that basic search will support this. 
nit : We will need to rewrite all existing unit tests (at least the ones touching basic search / other memory APIs)

aszlam:
This is ready for review now.  It is a "shallow" change, as most of the old FILTERS machinery is still in interpreter, despite the memory searcher being much more powerful... so the interpreter isn't more complicated after this; but now it can be simplified a lot.  After we retrain models with new spec, we can start refactoring/simplifying filters_helper, attribute_helper, etc.

kavyasrinet (approved):
This is a big change so I am in favor of landing this soon, but overall looks really good to me. Thanks for adding in the documentation, tests and updating basic search

Issue #471, opened 2021-06-08
Tracker for additional unit tests


kavyasrinet:
## Type of Issue

Select the type of issue:
- [ ] Bug report (to report a bug)
- [ ] Feature request (to request an additional feature)
- [x] Tracker (I am just using this as a tracker)
- [ ] Refactor request
- [ ] Documentation Ask

## Checklist

Starting this tracker for list of functions to add unit tests for. If you see any failures after the refactor, please add to this list (and the PR fixing the issue should add the test):
- [x] `build_random_shape` in craftassist's `default_behavior`
- [ ] test handle_dig (PR: https://github.com/facebookresearch/droidlet/pull/540)
- [ ] Interpreter:
   - [x] attribute_helper (PR: https://github.com/facebookresearch/droidlet/pull/510) 
      - [x] interpret_span_value
      - [x] maybe_specific_mem
      - [x] interpret_linear_extent
      - [x] interpret_task_info
    - [ ] comparator_helper.py
      - [ ] interpret_comparator
    - [ ] filter_helper.py 
    - [ ] get_memory_handler
    - [ ] interpreter_utils
    - [ ] location_helpers
    - [ ] reference_object_helpers
  - [ ] craftassist folder:
    - [ ]  block_helpers
    - [ ] dance.py
    - [x] default_behaviors
    - [ ] facing_helper
    - [ ] point_target
    - [ ] schematic_helper
    - [ ] size_words
    - [ ] spatial_reasoning
    - [ ] modify_helpers
  - [ ] locobot
    - [ ] dance
    - [ ] default_behaviors
    - [ ] facing_helper
    - [ ] point_target
    - [ ] get_memory_handler
    - [ ] spatial_reasoning
    - [ ]

Rebecca-Qian:
Semantic parser model training script was broken during refactor. In the previous setting, the training script was in the same directory as ML modules. In the refactor the imports were converted from file imports to module imports, and since the train script was invoked from `__main__`, this broke the training script imports.

It is strictly better to keep train/inference scripts and module classes separate, so I moved the scripts out into `tools/nsp_scripts/`. Now NSP is a standalone module.

Fix is in #485

Re: unit tests, I'm working on some basic integration tests. 🙂

snyxan:
Craftassist `perceive()` fn is failing due to parser/perception refactor. See this issue #497 and this PR #498 

We should add unittest for this part, tracked in this issue #499

kavyasrinet:
[...]

PR #485, opened 2021-06-10
[NSP] Refactor training script out of semantic parser module

Rebecca-Qian:
# Description

The training script for the semantic parser was in the same directory as the ML modules, which is bad design if we want to use NSP code as a standalone module. This is also functionally broken because you cannot import python modules in a script if that script also exists inside the modules path (called from `__main__`). I moved this and the inference script to `tools/nsp_scripts`. Also updated some other paths in inference code.

Fixes # (issue)

## Type of change

- [x] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

No changes to other components other than training needs to be run from `./tools/nsp_scripts/` now.

# Testing

# Checklist:


kavyasrinet:
@Rebecca-Qian Can we add unit tests to test this part of the code, in addition, in this PR ? The refactor didn't capture this part breaking and so that's a clear signal for adding tests for this.

Rebecca-Qian:
> @Rebecca-Qian Can we also add unit tests to test this part of the code, in this PR ? The refactor didn't capture this part breaking and so that's a clear signal for adding tests for this.

The training script broke, 

> @Rebecca-Qian Can we also add unit tests to test this part of the code, in this PR ? The refactor didn't capture this part breaking and so that's a clear signal for adding tests for this.

As discussed over chat:
The part that broke was training code, which does not touch production components. I do think it's important that we capture these breakages though, and this supports the need for integration tests.

Added a note to #471 and will also follow up with integration tests.

kavyasrinet (approved):

PR #486, opened 2021-06-10
registered basic db_write hook for dashboard timeline

am-zhang:
# Description

Added a basic hook to db_write in sql_memory.py to emit events to the dashboard. Also allows for optional logging from db_write to timeline_log.txt by entering `python craftassist.py --timeline_log` on startup. The hook is tentatively set to pick up VoxelObjects queries and the results as JSON. The latest event is displayed on the Timeline panel of the dashboard.

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After
Before:
Screen Shot 2021-06-10 at 2 41 33 PM

After:
Screen Shot 2021-06-11 at 4 32 39 PM

# Testing

# Checklist:


agents/craftassist/craftassist_agent.py
```diff
@@ -87,6 +87,14 @@ def __init__(self, opts):
         ]
         self.perceive_on_chat = True
 
+        # Add optional logging for timeline
+        if opts.log_timeline:
```
snyxan, line 91: This is redundant. If you just want to create this file, you can use 'a+' mode when open the file in `log_to_dashboard` fn.

agents/craftassist/craftassist_agent.py
```diff
@@ -87,6 +87,14 @@ def __init__(self, opts):
...
+            timeline_log = open("timeline_log.txt", "w")
```
snyxan, line 92: You should add some kind of id to the log file. Please refer to how to name agent log file: https://github.com/facebookresearch/droidlet/blob/amanda/timeline/agents/craftassist/craftassist_agent.py#L145

agents/craftassist/craftassist_agent.py
```diff
@@ -319,6 +327,15 @@ def add_self_memory_node(self):
...
         PlayerNode.create(self.memory, p, memid=self.memory.self_memid)
 
+    def log_to_dashboard(self, **kwargs):
+        """Emits the event to the dashboard and/or logs it in a file"""
+        result = kwargs['data']
+        self.agent_emit(result)
+        if self.opts.log_timeline:
+            timeline_log = open("timeline_log.txt", "a")
+            print(result, file=timeline_log)
+            timeline_log.close()
```
snyxan, line 337: You may want to keep the file open and only close it when the agent terminates.

snyxan (commented):
Thanks for creating this PR. Left a few comments.

droidlet/memory/sql_memory.py
```diff
@@ -997,6 +1002,11 @@ def db_write(self, query: str, *args) -> int:
         if self.on_delete_callback is not None and deleted:
             self.on_delete_callback(deleted)
         self._db_write("DELETE FROM Updates")
+        # a sample filter for emitting to the dashboard
+        if "VoxelObjects" in query:
+            hook_data = {"name" : "db_write", "time" : datetime.now().strftime("%H:%M:%S"), "query" : query, "args" : args, "result" : r}
```
snyxan, line 1007: Now the format of the data from db_write is query string, args list, etc. which is human-readable. You need a more structured format of data so that you can later use it to display on timeline (e.g. to show in a table, or support more fine-grained search). One way to do so is to convert the string to a key-value dict.

droidlet/memory/sql_memory.py
```diff
@@ -997,6 +1002,11 @@ def db_write(self, query: str, *args) -> int:
         if self.on_delete_callback is not None and deleted:
             self.on_delete_callback(deleted)
         self._db_write("DELETE FROM Updates")
+        # a sample filter for emitting to the dashboard
+        if "VoxelObjects" in query:
```
snyxan, line 1006: Move this filtering logic to `log_to_dashboard` fn instead of doing it here.

agents/craftassist/craftassist_agent.py
```diff
@@ -319,6 +327,15 @@ def add_self_memory_node(self):
             return
         PlayerNode.create(self.memory, p, memid=self.memory.self_memid)
 
+    def log_to_dashboard(self, **kwargs):
+        """Emits the event to the dashboard and/or logs it in a file"""
```
snyxan, line 331: Tip: You should check which function triggers this hook and process the data accordingly (you may not need it rn but will after we later log more events)

agents/craftassist/craftassist_agent.py
snyxan: Actually it's not redundant, you should open the file and get the handler here if `log_timeline` is set to true. Then you wouldn't need to open it every time in `log_to_dasboard` fn.

agents/craftassist/craftassist_agent.py
snyxan: As stated above^, you don't need to open it over and over here, just use the file handler you created during initialization.

agents/craftassist/craftassist_agent.py
```diff
@@ -87,6 +88,13 @@ def __init__(self, opts):
         ]
         self.perceive_on_chat = True
 
+        # Add optional logging for timeline
+        if opts.log_timeline:
+            self.timeline_log = open("timeline_log.{}.txt".format(self.name), "a+")
```
snyxan, line 93: nit: `timeline_log` -> `timeline_log_file`

agents/craftassist/craftassist_agent.py
snyxan: Assuming log_to_dashboard will be registered to multiple functions, you should do a switch here. A simple example using if-else (a cleaner alternative is lookup table):
```
fn_name = kwargs['name']
if fn_name is "db_write":
    do something
elif fn_name is "xxx":
    do something else
```

agents/craftassist/craftassist_agent.py
```diff
@@ -87,6 +88,13 @@ def __init__(self, opts):
         ]
         self.perceive_on_chat = True
 
+        # Add optional logging for timeline
+        if opts.log_timeline:
```
snyxan, line 92: Could you move this part to `loco_mc_agent.py`? similar to how to initialize db log file in `sql_memory.py` instead of `mc_memory.py` https://github.com/facebookresearch/droidlet/blob/amanda/timeline/droidlet/memory/sql_memory.py#L89

droidlet/memory/sql_memory.py
```diff
@@ -997,8 +1001,50 @@ def db_write(self, query: str, *args) -> int:
...
+            "table_name" : query_table, 
+            "operation" : query_operation, 
+            "args" : query_dict, 
+            "result" : r
+        }
+        hook_data = json.dumps(hook_data, default=str)
+        self.dispatch_signal.send(self.db_write, data=hook_data)
         return r
 
+    def parse_sql(self, query):
```
snyxan, line 1019: `parse_sql` and `format_query` should be moved to a separate file (`droidlet/memory/memory_util.py`). @aszlam  wdyt?

snyxan:
Could you update the screenshot with new data format after query parsing, thanks!

snyxan (approved):
LGTM. Ship it!

droidlet/memory/sql_memory.py
```diff
@@ -1153,3 +1169,9 @@ def safe_unpickle(self, bs):
         obj = pickle.loads(bs)
         self.reinstate_attrs(obj)
         return obj
+
+    def register_hook(self, receiver, sender):
+        """
+        allows for registering hooks using the event dispatcher
+        """
+        self.dispatch_signal.connect(receiver, sender)
```
soumith, line 1177: you should verify that the sender can only be one of `self.db_write`, `self.db_read`, etc.

You shouldn't be registering hooks for any sender via `memory.register_hook`

soumith (commented):
see inline comment

droidlet/memory/sql_memory.py
am-zhang: Will implement this right now, thanks for the feedback!

PR #506, opened 2021-06-19
backoff filter, use in interpreting dance filter

aszlam:
# Description

Fixes the "bug" that the agent does not dance when the command is "dance", even though parse is in gt.  this happened because the name of the dance is "dance", and that isn't found because there isn't a dance with that name (dances were not backing off to tags). 

this PR fixes this (and makes possible other backoffs more cleanly) by making a backoff filter that checks for the existence of memories from a sequence of Filters.  if any one Filter returns a memory, the BackoffFilter goes with that.

Fixes # (issue)

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

Dances will be found if they are just tagged according to the obj text, even when the parser returns a name search (like reference objects currently)

PR also makes a more general method of doing a backoff that should be used in other places (e.g. Reference objects and GetMemory)

# Testing

ran tests in game

# Checklist:

### DIFF
--- a/droidlet/interpreter/craftassist/dance.py
+++ b/droidlet/interpreter/craftassist/dance.py
@@ -59,11 +59,13 @@ head_bob = [
 
 def add_default_dances(memory):
     """Add dances to agent's memory"""
-    memory.add_dance(generate_sequential_move_fn(jump), name="jump")
+    memory.add_dance(generate_sequential_move_fn(jump), name="jump", tags=["jump"])
     memory.add_dance(
         generate_sequential_move_fn(konami_dance), name="konami dance", tags=["dance", "konami"]
     )
-    memory.add_dance(generate_sequential_move_fn(head_bob), name="head bob", tags=["dance"])
+    memory.add_dance(
+        generate_sequential_move_fn(head_bob), name="head bob", tags=["dance", "head bob"]
+    )
 
 
 def generate_sequential_move_fn(sequence):
--- a/droidlet/interpreter/filter_helper.py
+++ b/droidlet/interpreter/filter_helper.py
@@ -13,6 +13,7 @@ from droidlet.memory.memory_filters import (
     CountTransform,
     ExtremeValueMemorySelector,
     RandomMemorySelector,
+    BackoffFilter,
 )
 from droidlet.base_util import number_from_span
 from ..shared_data_structs import ErrorWithResponse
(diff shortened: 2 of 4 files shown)


droidlet/interpreter/craftassist/dance.py
```diff
@@ -59,11 +59,13 @@
 
 def add_default_dances(memory):
     """Add dances to agent's memory"""
-    memory.add_dance(generate_sequential_move_fn(jump), name="jump")
+    memory.add_dance(generate_sequential_move_fn(jump), name="jump", tags=["jump"])
```
kavyasrinet, line 62: and another tag could be "dance" here, right ?

droidlet/interpreter/filter_helper.py
```diff
@@ -230,13 +231,27 @@ def interpret_task_filter(interpreter, speaker, filters_d, get_all=False):
...
+
     search_data = {}
     triples = []
+    triple_filter = None
     for t in filters_d.get("triples"):
         triples.append(t)
-    search_data["base_table"] = "Dances"
-    search_data["triples"] = triples
-    F.append(BasicFilter(interpreter.memory, search_data))
+    if triples:
```
kavyasrinet, line 240: not sure I understand why we have this snippet here. Looks like triples isn't assigned anything here, we won't even enter the code block, right ?

droidlet/lowlevel/minecraft/cuberite_process.py
```diff
@@ -121,7 +126,7 @@ def atexit(self):
     import argparse
 
     parser = argparse.ArgumentParser()
-    parser.add_argument("--config", default="diverse_world")
+    parser.add_argument("--config", default="flat_world")
```
kavyasrinet, line 129: thanks for changing this!

droidlet/memory/memory_filters.py
```diff
@@ -590,6 +590,39 @@ def _selfstr(self):
         return "Basic: (" + str(self.search_data) + ")"
 
 
+class BackoffFilter(MemoryFilter):
+    """
+    runs a sequence of Filters, passed into the __init__.  if nothing is returned from the
```
kavyasrinet, line 595: Can we rephrase this to something like - goes over the list of filters in __init__ and returns the first not None (memids, vals) tuple ?

kavyasrinet (changes requested):

droidlet/interpreter/craftassist/dance.py
aszlam: I think not, we don't want a single jump to be returned when a player asks the agent to dance ?

droidlet/interpreter/filter_helper.py
aszlam: the code block is switched on filters_d.get("triples"), and the triples start empty.  if the filter dict has triples, they will be appended to the list named "triples", and that list will be converted to a BasicSearch Filter.  I think the code is correct as it is, but more wordy than it needs to be (could just use filters_d["triples"] in line 242; but I just wanted to match the structure of the block at 249

droidlet/interpreter/craftassist/dance.py
kavyasrinet: Jumps are dances though, right ?

droidlet/interpreter/filter_helper.py
kavyasrinet: Ah nvm I missed this piece:
```
for t in filters_d.get("triples"):
        triples.append(t)
```
where we change triples

droidlet/interpreter/craftassist/dance.py
aszlam: they are Dances in the sense of the Dance Task (a sequence of movements where the sequence is more important than the destination) but not in the standard common parlance.  I  am using this tag as a thing that can be searched over to return a dance when no particular dance is specified; but I think in that situation the person commanding the agent wants the agent to dance in the common sense

kavyasrinet (approved):

PR #525, opened 2021-06-25
Jhan/interaction logger separate files

jhan6501:
# Description
Interaction files are saved into separate files for different sessions. This prevents newer sessions from overwriting previous sessions when uploading the json file to AWS S3 bucket

Fixes # (issue)

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

image

# Testing

# Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,7 +44,7 @@ incoming_chat.txt
 locobot/test/test_assets/perception_handlers/
 locobot/agent/dashboard_data.db*
 nsp_outputs.csv
-agents/locobot/interaction_loggings.json
+agents/locobot/interaction_loggings_*
 agents/locobot/models
 agents/locobot/datasets
 agents/locobot/dashboard_data.db
--- a/.gitignore
+++ b/.gitignore
@@ -44,7 +44,9 @@ incoming_chat.txt
 locobot/test/test_assets/perception_handlers/
 locobot/agent/dashboard_data.db*
 nsp_outputs.csv
-agents/locobot/interaction_loggings_*
+agents/locobot/*.json
+agents/locobot/*.csv
+agents.locobot/*.log
 agents/locobot/models
 agents/locobot/datasets
 agents/locobot/dashboard_data.db
--- a/agents/locobot/locobot_agent.py
+++ b/agents/locobot/locobot_agent.py
@@ -125,30 +125,19 @@ class LocobotAgent(LocoMCAgent):
             self.dashboard_memory["objects"] = objects
             sio.emit("updateState", {"memory": self.dashboard_memory})
 
-        @sio.on("create loggings file")
-        def create_loggings_file(sid, id):
-            print ("HAI WE ARE CREATING LOGGINGS FILE IN LOCOBOT_AGNET.PY")
-            print ("HAI WE ARE CREATING LOGGINGS FILE IN LOCOBOT_AGNET.PY")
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -44,7 +44,7 @@ incoming_chat.txt
 locobot/test/test_assets/perception_handlers/
 locobot/agent/dashboard_data.db*
 nsp_outputs.csv
-agents/locobot/interaction_loggings.json
+agents/locobot/interaction_loggings_*
```
Rebecca-Qian, line 47: still think you should add paths like *.csv, *.json, or *.log, because other components can write logs too that you don't want committed

agents/locobot/locobot_agent.py
```diff
@@ -81,6 +82,7 @@ def __init__(self, opts, name="Locobot"):
         self.init_event_handlers()
         # list of (prob, default function) pairs
         self.visible_defaults = [(1.0, default_behaviors.explore)]
+        self.interactionLogger = None
```
Rebecca-Qian, line 85: nit: use snake case

agents/locobot/locobot_agent.py
```diff
@@ -123,6 +125,15 @@ def objects_in_memory(sid):
             self.dashboard_memory["objects"] = objects
             sio.emit("updateState", {"memory": self.dashboard_memory})
 
+        @sio.on("store session id")
+        def create_loggings_object(sid, id):
+            filePath = "interaction_loggings_" + id + ".json"
+            self.interactionLogger = InteractionLogger(filePath)
```
Rebecca-Qian, line 131: nit (and this applies to this whole file): snake case for python

see https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

droidlet/perception/semantic_parsing/utils/interaction_logger.py
```diff
@@ -13,6 +14,15 @@ def __init__(self, filepath):
...
         self.log_filepath = filepath
+        self.createdFile = False
+
+
+    def create_loggings_file(self, filepath):
+        self.createdFile = True
+
+        if not os.path.isfile(self.log_filepath):
+            with open(self.log_filepath, 'w') as f:
+                json.dump([], f, ensure_ascii=False, indent=4)
```
Rebecca-Qian, line 25: You're checking that the log file does not exist, then create an empty file? I remember we discussed this, what I meant was, if the file exists, you load it into memory and append interactions. If not, write to it.

But I also think it's fine to just overwrite this file with the assumption that's you're saving one log file single session.

jhan6501: This function gets called in logInteraction. Basically, if the file isn't created, create the file with an empty json list. This is important because if the file doesn't exist, reading the non-existent file as a json will crash. I will add comments/rewrite this method to make this more clear.

Rebecca-Qian: > This function gets called in logInteraction. Basically, if the file isn't created, create the file with an empty json list. This is important because if the file doesn't exist, reading the non-existent file as a json will crash. I will add comments/rewrite this method to make this more clear.

What I mean is, you don't need to create an empty file here, you can just create it when writing. Anyway, this is not a big deal, if you want to keep it this way you can.

Rebecca-Qian (approved):

PR #529, opened 2021-06-29
feed info in StateManager state

hollisma:
# Description

Stored the rgb image, depth image, masks, and pose in the StateManager state. The state for the current frame and the last different frame are both stored. Right now, the state is only being maintained and is not used anywhere. 

This PR is to help unblock @jhan6501 so that he can use the state to update mobile annotations. Not all the properties are needed for Jerry, but I also plan on using the state for label propagation in the future so I included the other properties to lower merge conflict costs. 

## Type of change

- [x] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Testing

Tested locally and confirmed that prevFeedState stores the state from the most recent frame that is different from the current frame.

### DIFF
--- a/droidlet/dashboard/web/src/StateManager.js
+++ b/droidlet/dashboard/web/src/StateManager.js
@@ -104,6 +104,20 @@ class StateManager {
     this.setUrl(url);
 
     this.fps_time = performance.now();
+
+    // Assumes that all socket events for a frame are received before the next frame
+    this.curFeedState = {
+      rgbImg: null, 
+      depth: null, 
+      masks: null,
+      pose: null,
+    }
+    this.prevFeedState = {
+      rgbImg: null, 
+      depth: null, 
+      masks: null,
+      pose: null,
+    }
   }
 
   setDefaultUrl() {
@@ -398,6 +412,9 @@ class StateManager {
         }
       }
     });
+    if (!this.curFeedState.rgbImg) {
+      this.curFeedState.rgbImg = res
+    }
   }
 
   processDepth(res) {
@@ -413,6 +430,9 @@ class StateManager {
         }
       }
     });
+    if (!this.curFeedState.depth) {
+      this.curFeedState.depth = res
+    }
   }
 
   processRGBDepth(res) {
@@ -441,6 +461,9 @@ class StateManager {
         });
       }
     });
+    if (!this.curFeedState.masks) {
+      this.curFeedState.masks = res.objects.map(o => o.mask)
+    }
   }
 
   processHumans(res) {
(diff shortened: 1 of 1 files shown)


jhan6501:
Looks good, but in ObjectFixup there is {this.state.objects} Is objects being stored in these feed states anywhere?

hollisma:
> Looks good, but in ObjectFixup there is {this.state.objects} Is objects being stored in these feed states anywhere?

Yes it's being stored as this.curFeedState.masks. In processObjects, the objects are passed to LiveObjects which then passes it down until it reaches ObjectFixup. Right after passing the objects, processObjects then sets the state of StateManager. So to resolve to mobile bug, we can add a check in ObjectFixup that sees if there are objects in its state, and if not, it gets the objects from the StateManager.

jhan6501 (approved):
lgtm

kavyasrinet:
Also looks good to me

droidlet/dashboard/web/src/StateManager.js
```diff
@@ -472,6 +495,30 @@ class StateManager {
         });
       }
     });
+
+    // Update feed state here because it's received first
```
soumith, line 499: what do you mean by it's received first?

hollisma: This is where the first socket event from the backend to the StateManager is processed

soumith: we might reorder the backend events over time. so can you rewrite the logic without this assumption?

hollisma: Yep will do!

droidlet/dashboard/web/src/StateManager.js
```diff
@@ -472,6 +495,30 @@ class StateManager {
...
+      res.yaw !== this.curFeedState.pose.yaw)) {
+      this.prevFeedState = this.curFeedState
+      this.curFeedState = {
+        rgbImg: null, 
+        depthOrg: null, 
+        masks: null, 
+        pose: {
+          x: res.x, 
+          y: res.y, 
+          yaw: res.yaw,
```
jhan6501, line 518: Why does rgbImg, depthOrg, and masks all being reset to null? wouldn't that override previous data?

droidlet/dashboard/web/src/StateManager.js
hollisma: @soumith In order to make the logic order-independent, we would need to check if the mask arrays are equal, which could be expensive. Plus, the data is sent over a few times every second, so the current logic will work even if the order is different as long as the data being sent does not change extremely often. I think it's still a good idea to change the logic but wanted to get your input on how to compare the large mask arrays. 

And the reasoning behind the large array comparisons is to see if we need to update the previous state.

soumith: send an incrementing `tick` integer with every message that is the same across the message in a given timestamp

soumith: maybe something like `self.count` of the agent?

soumith: https://github.com/facebookresearch/droidlet/blob/main/agents/core.py#L32

hollisma: The array comparisons turned out to be fast so I just did that. The 'tick' method might be cleaner and more efficient but would take longer to implement. Let me know if you think it's a good idea to prioritize changing the implementation to using ticks.

droidlet/dashboard/web/src/StateManager.js
hollisma: This should be solved with the most recent commit. The only time the properties can be null should only be at the start.

soumith (approved):

PR #537, opened 2021-06-30
[polymetis] Fix timestamp representation

exhaustin:
# Description

Naively converting the timestamp to a float would result in loss of precision, causing time differences to be incorrectly calculated.

- Change the tensor representation of a timestamp to a int32 tensor of size 2, with the first element representing seconds and the second representing nanoseconds.
- Add helper functions to easily obtain time differences between two timestamps.

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

Before this change, timestamps would get converted to the same float value due to [loss of significance](https://en.wikipedia.org/wiki/Loss_of_significance), and thus trying to get differences between timestamps would always result in 0 seconds.

After the change, helper functions return accurate time differences between two timestamps.

# Testing

Add integration test `test/scripts/3_timestamps.py`

# Checklist:

### DIFF
--- a/polymetis/polymetis/python/torchcontrol/utils/time_utils.py
+++ b/polymetis/polymetis/python/torchcontrol/utils/time_utils.py
@@ -1,3 +1,7 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
 import torch
 
 
--- a/polymetis/polymetis/python/torchcontrol/utils/time_utils.py
+++ b/polymetis/polymetis/python/torchcontrol/utils/time_utils.py
@@ -2,16 +2,53 @@ import torch
 
 
 def timestamp_subtract(ts1, ts2):
+    """ Subtracts two timestamps
+
+    Timestamps are represented as a torch.Tensor of shape (2,), where the first element 
+    corresponds to seconds and the second element corresponds to nanoseconds.
+
+    Args:
+        ts1: First timestamp
+        ts2: Second timestamp
+
+    Outputs:
+        Result of ts1 - ts2 represented as a timestamp
+    """
+    
     assert ts1.shape == torch.Size([2])
     assert ts2.shape == torch.Size([2])
     return ts1 - ts2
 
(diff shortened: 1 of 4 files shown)


polymetis/polymetis/python/torchcontrol/utils/time_utils.py
```diff
@@ -0,0 +1,54 @@
+import torch
```
1heart, line 1:
```suggestion
# Copyright (c) Facebook, Inc. and its affiliates.
#
# This source code is licensed under the MIT license found in the
# LICENSE file in the root directory of this source tree.
import torch
```

polymetis/tests/scripts/3_timestamps.py
```diff
@@ -0,0 +1,61 @@
...
+        self.i = 0
+        self.ts_prev = torch.zeros(2).to(torch.int32)
+
+    def forward(self, state_dict: Dict[str, torch.Tensor]):
+        ts = state_dict["timestamp"]
+
+        # Check timestamp
+        if self.i > self.warmup_steps:
+            t_diff = toco.utils.timestamp_diff_seconds(ts, self.ts_prev)
+            assert torch.allclose(t_diff, self.dt, atol=1e-3)  # millisecond accuracy
```
1heart, line 38: Is this test dependent on the machine it's run on? E.g. do we expect this test to become flaky if CircleCI's VMs get slow for whatever reason?

1heart (approved):
Minor nits; otherwise looks good, thanks! This was harder than I expected timestamps to be :D

polymetis/tests/scripts/3_timestamps.py
exhaustin: Yes which is why I'm setting a very large tolerance.
An alternative is that we can also test if t_diff > 0, but the current test would also ensure that the frequency provided in the metadata from the simulation client is correct so I went for it.
I'm fine with either, thoughts?

1heart: sgtm

PR #551, opened 2021-07-02
dialogue and interpreter hooks for dashboard timeline

### LINKED ISSUE #550: move `event` system to have a centralized Signal object and use `string` keys
Currently, the way we register hooks on events is something like [this](https://github.com/facebookresearch/droidlet/blob/06f9e09e7c45d2326285ff556506cc45e5c2c243/agents/loco_mc_agent.py#L67)

```python
self.register_hook(self.log_to_dashboard, self.perceive)
```

Here, we are registering a closure called `self.log_to_dashboard` to be executed whenever `self.perceive` is finished.

The way we do this is to create a `dispatch.Signal()` object, and then register the `receiver` / `sender` onto this object.

```python
    self._dispatch_signal = dispatch.Signal()
    def register_hook(self, receiver, sender):
         self._dispatch_signal.connect(receiver, sender)
```
Then, in `perceive`, we call something like this:

```python
self._dispatch_signal.send(self.perceive, data=hook_data)
```

However, this relies on two things:

1. the hook to be registered on a non-ephemeral object, ahead of time (for example if the object is deleted and created again, the hook is lost)
2. the hook relies on a function pointer of a function in the object, i.e. `self.perceive`

This is not really perfect for all kind of dispatch hooks we want across the codebase. What we want is a more straightfoward pub/sub mechanism, where any function anywhere should be able to subscribe to a given event, and any other line of code anywhere should be able to emit that event with some data.

My proposal is that we do the following:

1. We use the `dispatch.Signal` object already created [here](https://github.com/facebookresearch/droidlet/blob/06f9e09e7c45d2326285ff556506cc45e5c2c243/droidlet/event/__init__.py#L11), instead of creating one everytime we need dispatchers. So, you can use it like `from droidlet.event import dispatch`
2. Add a conditional to the `_make_id` here to special-case strings to be their own content, rather than their pointers: https://github.com/facebookresearch/droidlet/blob/06f9e09e7c45d2326285ff556506cc45e5c2c243/droidlet/event/dispatcher.py#L7
  - the change would look something like `if isinstance(target, str): return target`
  - this would basically enable us to register string keys for dispatch

With these two changes, then we can start directly using the global dispatch object and use string-keys to register hooks.

```python
from droidlet.event import dispatch

# no longer need a register_hook

# registering a hook
dispatch.connect(self.log_to_dashboard, "agent_perceive")


# in perceive
dispatch.send("agent_perceive", data=hook_data)
```

am-zhang:
# Description

This PR adds hooks for dialogue and interpreter for data to be sent to the dashboard. It modifies the existing event dispatch system to work with ephemeral objects. It also adds styling to the timeline component in order to display the three types of events.

Run by entering `python craftassist_agent.py --log_timeline --enable_timeline`.

Fixes #550 

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

Screen Shot 2021-07-01 at 6 57 58 PM

# Testing

Tested by running the dashboard with CraftAssist.

# Checklist:


soumith:
I reviewed commit 34f26aa which looks good to me

kavyasrinet (approved):
Looks good to me after the dispatcher changes.

PR #560, opened 2021-07-08
Fix occasionally flaky locobot_agent shutdown logic

soumith:
The comments in the code describe the change well

### DIFF
--- a/agents/locobot/locobot_agent.py
+++ b/agents/locobot/locobot_agent.py
@@ -240,7 +240,16 @@ class LocobotAgent(LocoMCAgent):
 
     def shutdown(self):
         self._shutdown = True
-        self.perception_modules["vision"].vprocess_shutdown.set()
+        try:
+            self.perception_modules["vision"].vprocess_shutdown.set()
+        except:
+            """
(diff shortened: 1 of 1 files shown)


agents/locobot/locobot_agent.py
```diff
@@ -240,7 +240,16 @@ def task_step(self, sleep_time=0.0):
...
+        try:
+            self.perception_modules["vision"].vprocess_shutdown.set()
+        except:
+            """
+            the try/except is there in the event that
+            self.perception_modules["vision"] has either:
+            1. not been fully started yet
+            2. already crashed / shutdown due to other effects
+            """
+            pass
```
Rebecca-Qian, line 252: So if this process crashes because it has not been fully started, we ignore?

anuragprat1k (approved):

agents/locobot/locobot_agent.py
soumith: yea, this stuff was there just to do a "clean" exit anyways. regardless of this line, the process is a "daemon" process, so it would stop when the main process stops.

PR #569, opened 2021-07-09
[polymetis] Automatic recovery for the Franka client

exhaustin:
# Description

Enable the Franka Panda client to attempt to recover automatically from an error instead of exiting.

### Changes:
- Wrap command that runs the robot within a loop that attempts to recover using `franka::Robot::automaticErrorRecovery` when an error occurs.
- Update PolymetisServer to re-launch a default controller instead of exiting when a stale robot client is detected.
- Update test to reflect the external termination service that is now one step faster due to changes in PolymetisServer.

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

Before the changes implemented, the franka client would run to a halt whenever an error is encountered, which is usually due to a safety constraint violation.
Screen Shot 2021-07-09 at 3 13 11 PM

After the changes, the client now attempts to reconnect to the robot and return to an operational state by launching a default controller from the new state.
Screen Shot 2021-07-09 at 3 11 11 PM

# Testing

Tested on hardware.

# Checklist:

### DIFF
--- a/polymetis/polymetis/src/polymetis_server.cpp
+++ b/polymetis/polymetis/src/polymetis_server.cpp
@@ -148,9 +148,21 @@ PolymetisControllerServerImpl::ControlUpdate(ServerContext *context,
     custom_controller_context_.status = TERMINATING;
   }
 
-  // Lock to prevent external termination during controller selection, which
-  // might cause loading of a uninitialized default controller
-  service_mtx_.lock();
+  // Parse robot state
+  auto timestamp_msg = robot_state->timestamp();
+  rs_timestamp_[0] = timestamp_msg.seconds();
+  rs_timestamp_[1] = timestamp_msg.nanos();
+  for (int i = 0; i < num_dofs_; i++) {
+    rs_joint_positions_[i] = robot_state->joint_positions(i);
+    rs_joint_velocities_[i] = robot_state->joint_velocities(i);
+    rs_motor_torques_measured_[i] = robot_state->motor_torques_measured(i);
+    rs_motor_torques_external_[i] = robot_state->motor_torques_external(i);
+  }
+
+  // Lock to prevent 1) controller updates while controller is running; 2)
+  // external termination during controller selection, which might cause loading
+  // of a uninitialized default controller
+  custom_controller_context_.controller_mtx.lock();
 
   // Update episode markers
   if (custom_controller_context_.status == READY) {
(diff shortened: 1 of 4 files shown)


polymetis/polymetis/src/polymetis_server.cpp
```diff
@@ -163,14 +185,6 @@ PolymetisControllerServerImpl::ControlUpdate(ServerContext *context,
     rs_motor_torques_external_[i] = robot_state->motor_torques_external(i);
   }
 
-  // Select controller
```
1heart, line 166: Why is it necessary to move this above parsing the robot state? Just trying to understand what's happening

polymetis/polymetis/tests/cpp/test_server.cpp
```diff
@@ -124,7 +124,7 @@ TEST_F(ServiceTest, SetController) {
                                       dummy_robot_state_, new TorqueCommand)
                       .ok());
 
-      if (i == 1) {
+      if (i == 2) {
```
1heart, line 127: I'm slightly suspicious of the changes in `polymetis_server.cpp` because it causes this subtle off-by-1 here... My summary of the effect is that cleaning up a "TERMINATING" controller occurs in the _next_ call to ControlUpdate. Any thoughts on potentially dangerous downstream effects of the changes?

1heart (approved):
I'm glad that resetting through software proved to be simple!

polymetis/polymetis/tests/cpp/test_server.cpp
exhaustin: The previous design cleans up a "TERMINATING" controller at the end of the current timestep, which includes the resetting of a the default controller. The problem with this is that if the controller status changes to TERMINATING before the controller selection step (either due to a untimely TerminateController call from the user, or due to stale controller state caused by a automatic restart), the controller would select and load a default controller that has not been reset to the current pose, which would have safety concerns.

Thus my solution is to process the TERMINATING controller _right_ before the controller selection step so that the default controller is always initialized when newly loaded. This is done by moving the step for cleaning up TERMINATING controllers up to the beginning of the loop, and also moving the controller selection step up to be right next to it. Could also parse the robot state before the two steps in question, it doesn't really matter.

The off-by-1 phenomena in the test is actually caused by the termination being one step __faster__. As you can see I increased the number of steps the controller was allowed to run by 1, while leaving the expected episode length untouched. This is because the ControlUpdate call after the TerminateController now finalizes the episode at the beginning of the loop, instead of loading & running the default controller and including it in the episode. I believe this is an improvement.

1heart: > The problem with this is that if the controller status changes to TERMINATING before the controller selection step (either due to a untimely TerminateController call from the user, or due to stale controller state caused by a automatic restart), the controller would select and load a default controller that has not been reset to the current pose, which would have safety concerns.

This is definitely a safety issue, so thanks for catching it! From what I can tell, we don't actually guarantee thread-safety from untimely TerminateController calls simply by processing the TERMINATING controller right before the controller selection -- we'd need a mutex for that. In fact, we already have a mutex to make controller switching safe (`service_mtx_`).

I think the fundamental issue is that switching to a default controller is not thread safe in our implementation. How about we eliminate `TERMINATING` as a state in our controller finite-state-machine, and instead have a method which switches to & resets the default controller, which is protected by `service_mtx_`?

exhaustin: I think using a mutex makes perfect sense here. 
However both internal and external terminations need to go through the same process of saving the episode indices in addition to resetting the default controller, which might cause race conditions with the controller loop.
Do you think it should be sufficient to simply protect the termination handling and controller selection part (the two blocks of code that I moved) with `service_mtx_`?

PR #575, opened 2021-07-13
Hollis/label prop integration

hollisma:
# Description

Implemented @anuragprat1k's label propagation script into the dashboard. This PR allows users to use the annotation tool to correct masks and propagate those masks to future frames. 

The corrected masks as well as the current and previous images, depth maps, and base pose data are sent to the backend to be processed and used as inputs into the label propagation script. The script then returns the new mask locations, which are rendered on the object detector video feed. Masks from the detector are displayed with white borders whereas annotated / propagated masks are displayed with black borders. 

More information in this [design doc](https://www.notion.so/hollism/Label-prop-design-4579c7d6892d4bd0a5f3342804e159f5)

Some limitations: 
* Propagated labels often have mask small masks (3 points) that can probably be taken out
* The colors of masks between frames is not consistent -- to make them consistent, we would just need to attach a color property to the objects
* Label propagation currently happens every time annotations are saved. Could make a toggle for this

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

Before annotating: 
image

After annotating: 
(masks are slightly diff because we cap the number of points displayed on the annotator tool)
image

After moving after annotating: 
The previous green / red masks are now blue / orange. Also, a reason why the propagated masks might look bad is because the input masks are bad. 
image

# Testing

I tested the changes by going through example user flows of annotating objects and moving the agent around to see the propagated labels. 

Would love to test this on simpler objects since these chairs and tables overlap and are difficult to fully mask manually. 

# Checklist:


agents/loco_mc_agent.py
```diff
@@ -110,6 +115,65 @@ def save_object_annotation_to_db(sid, postData):
             logging.debug("in save_object_annotation_to_db, got postData: %r" % (postData))
             saveObjectAnnotationsToDb(self.conn, postData)
 
+        @sio.on("label_propagation")
+        def label_propagation(sid, postData):
```
kavyasrinet, line 119: I am assuming this is only specific to `locobot_agent.py` ? If so, I'd recommend adding this event handler to : `agents/locobot/locobot_agent.py` . 
We should try and add socket events that are common to all droidlet agents to this file.

droidlet/dashboard/web/src/StateManager.js
```diff
@@ -105,19 +109,24 @@ class StateManager {
 
     // Assumes that all socket events for a frame are received before the next frame
     this.curFeedState = {
-      rgbImg: null,
-      depth: null,
-      masks: null,
+      rgbImg: null, 
+      depth: null, 
+      objects: null, // Can be changed by annotation tool
+      orgObjects: null, // Original objects sent from backend
```
kavyasrinet, line 115: nit: `orgObjects` -> `origObjects`

kavyasrinet (changes requested):

agents/loco_mc_agent.py
hollisma: From my understanding, we want label prop to run for any agent that has a detector. Although if `loco_mc_agent.py` is meant for only locobot and the mc agent, then moving this socket event to `locobot_agent.py` would make sense. Can I confirm that `loco_mc_agent.py` is only for these two agents? @anuragprat1k

agents/locobot/locobot_agent.py
```diff
@@ -129,6 +134,64 @@ def objects_in_memory(sid):
...
+                depth_imgs.append(depth_scaled)
+            src_depth = np.array(depth_imgs[0])
+            cur_depth = np.array(depth_imgs[1])
+
+            # Convert mask points to mask maps then combine them
+            src_label = np.zeros((height, width)).astype(int)
+            for n, o in enumerate(postData["prevObjects"]): 
+                poly = Polygons(o["mask"])
+                bitmap = poly.mask(height, width)
+                # TODO probably a cleaner way to do this with numpy arrays
```
anuragprat1k, line 166: you can replace this for loop with
`src_label[bitmap == True] = n+1`

droidlet/dashboard/web/src/StateManager.js
```diff
@@ -380,6 +393,117 @@ class StateManager {
...
+    this.stateProcessed.depth = true;
+    this.stateProcessed.objects = true;
+    this.stateProcessed.pose = true;
+  }
+
+  labelPropagationReturn(res) {
+    this.refs.forEach((ref) => {
+      if (ref instanceof LiveObjects) {
+        for (let i = 0; i < res.length; i++) {
+          // Get rid of masks with <3 points
```
anuragprat1k, line 474: I see this (getting rid masks with <3 points) done here and in `onObjectAnnotationSave` above. Why are both needed?

anuragprat1k (commented):
For the scenario when a mask has both the detector mask and an annotated mask, what happens when there is an overlap between the two? How is that rendered?

hollisma:
> For the scenario when a mask has both the detector mask and an annotated mask, what happens when there is an overlap between the two? How is that rendered?

Both masks are rendered. I think we decided to let the user decide which mask they want to propagate and they can delete any masks they don't want.

droidlet/dashboard/web/src/StateManager.js
hollisma: One checks points generated by the label prop script and the other checks points generated by the annotator tool. Although we already have a check in the annotator tool, so I'll get rid of one of these checks.

anuragprat1k:
> > For the scenario when a mask has both the detector mask and an annotated mask, what happens when there is an overlap between the two? How is that rendered?
> 
> Both masks are rendered. I think we decided to let the user decide which mask they want to propagate and they can delete any masks they don't want.

We don't want the user to choose in every frame which mask to propagate and by default only want to propagate the annotated mask - once they enable label prop, their annotated mask keeps getting propagated (they can choose to keep correcting it in between when the propagation gets bad). Even if both the detector and annotated masks are shown, is the default propagation of the annotated mask achieved here?

hollisma:
> > > For the scenario when a mask has both the detector mask and an annotated mask, what happens when there is an overlap between the two? How is that rendered?
> > 
> > 
> > Both masks are rendered. I think we decided to let the user decide which mask they want to propagate and they can delete any masks they don't want.
> 
> We don't want the user to choose in every frame which mask to propagate and by default only want to propagate the annotated mask - once they enable label prop, their annotated mask keeps getting propagated (they can choose to keep correcting it in between when the propagation gets bad). Even if both the detector and annotated masks are shown, is the default propagation of the annotated mask achieved here?

Yes, annotated masks are always fed as inputs into label prop whereas detected masks are never propagated (unless the user corrects them).

anuragprat1k (approved):

kavyasrinet (approved):
Nit: 
I see an addition : `import droidlet.event.dispatcher as dispatch` in `loco_mc_agent` but `dispatch` not being used in the changeset, master has [this](https://github.com/facebookresearch/droidlet/blob/main/agents/loco_mc_agent.py#L14) 
Can you double check this and remove the unused import ?

Other than that LGTM

agents/locobot/locobot_agent.py
hollisma: It turns out `bitmap` is not a np array and cannot be easily converted into one (afaik), so I'll leave this as is for now.

PR #591, opened 2021-07-16
[polymetis] Habitat Manipulator environment

1heart:
# Description

Changes:

- Create a Polymetis environment around [habitat-sim](https://github.com/facebookresearch/habitat-sim)
- Use Habitat v2's ability to create articulated objects from URDFs & control them
- Update CircleCI with unit tests/habitat-sim

Note: habitat-sim's physics simulation dynamics (based on bullet3) is subtly different than PyBullet's, since PyBullet wraps the bullet3 commands in a slightly-higher-level API. The result is that there needs to be significant tuning for the physics parameters (tunings, etc) and the end result still is quite different: requires different (lower) gains, and the Pinocchio-based gravity compensation torques don't hold the arm still.

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

Running `launch_robot.py robot_client=habitat_sim use_real_time=false habitat_dir=/path/to/habitat-sim/ gui=true` launches a Franka in an apartment:

image

# Testing

Added unit tests.

# Checklist:


polymetis/polymetis/python/polysim/envs/habitat_manipulator.py
```diff
@@ -0,0 +1,286 @@
...
+        return (
+            self.robot_model.inverse_dynamics(
+                torch.tensor(pos),
+                torch.tensor(vel),
+                torch.tensor([0 for _ in range(self.n_dofs)]),
+            )
+            .detach()
+            .cpu()
+            .numpy()
+        )
```
exhaustin, line 247: By having the current velocity as input you are also compensating for the Coriolis term in addition to gravitational forces. This differs from the [implementation in bullet_manipulator.py](https://github.com/facebookresearch/droidlet/blob/59cc99714703831e5c8c10dc5c4d9ac2c633c4a2/polymetis/polymetis/python/polysim/envs/bullet_manipulator.py#L161), which only compensates for gravity.
Is this intentional due to discrepancies between Pybullet and Habitat?

polymetis/polymetis/python/polysim/envs/habitat_manipulator.py
```diff
@@ -0,0 +1,286 @@
...
+    """Moves robot to reasonable transformation relative to agent."""
+    if local_base_pos is None:
+        local_base_pos = np.array([0.0, -1.1, -2.0])
+    # place the robot root state relative to the agent
+    agent_transform = sim.agents[0].scene_node.transformation_matrix()
+    base_transform = mn.Matrix4.rotation(
+        mn.Rad(angle_correction), mn.Vector3(1.0, 0, 0)
+    )
+    base_transform.translation = agent_transform.transform_point(local_base_pos)
+    robot.transformation = base_transform
```
exhaustin, line 121: Are the values for agent state and base position specific to the Franka?
If so maybe they should be set from the configs?

exhaustin (approved):
Looks good in general, few minor concerns in comments.
Approving first as no changes would affect existing features.

Also, I'm unable to get habitat to run on my NUC as magnum complains that `libOpenGL.so.0` cannot be found. Resolving this right now but this might indicate that more dependencies need to be specified.

polymetis/polymetis/python/polysim/envs/habitat_manipulator.py
1heart: Nope, just fixed, thanks!

1heart: Nope, just fixed, thanks! Still doesn't do the right thing if we just input gravcomp torques, though; requires some more digging.

polymetis/polymetis/python/polysim/envs/habitat_manipulator.py
1heart: While not specific to the Franka, they are specific to the scene. I put them as optional parameters in HabitatManipulator but did not add them to the yaml configs, since this should probably be tweaked in detail when someone iterates on specific Habitat scenes.

1heart:
The libopenGl issue is probably [the freeglut3](https://github.com/facebookresearch/droidlet/blob/main/polymetis/docker/ubuntu18/Dockerfile#L18) packages installed as part of the base OS in the CI Dockerfile?

PR #601, opened 2021-07-22
Hollis/retrain detector

hollisma:
# Description

* This PR lets users save annotations and use those annotations to retrain the detector. 
* The retraining is centered around saving and reading files and models from a new folder called “/annotation_data/”. 
* Mechanics
    * We **save the annotations** and category/properties to the disk each time label propagation is run. 
    * When the user clicks **“Save,”** the files are read from and put into a COCO-formatted json file. The categories and properties are also saved into their own json files (used in the Detector). 
    * When the user clicks **“Retrain,”** the COCO json file is split into test and training files, which are then used to train a detectron2 model from scratch. We also evaluate the model here. 
    * Once the retraining and evaluation are done, the **evaluation metrics** are displayed on the frontend and a button appears that lets the user **“Switch”** to the new model. 
    * On switching, we feed the new folder path to a **new Perception instance** and connect it to the frontend. 
* Other notes: 
  * @soumith and I decided that we should have code-based category and property files instead of the pickle files that are currently used. This will require changing the tar-gz folder downloaded when the —dev flag is not used, and removing things.pickle and prop.pickle from the tar-gz. There is also some hardcoded code in detector.py to process the pickle files, but this can be removed once the pickle files are gone. 
  * I added a panel on the right called "Retrainer" that allows users to tune hyperparameters. 

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

Before buttons: 
image

After buttons: 
image

Initial annotation_data/: 
image

Annotation_data/ after 3 label prop calculations: 
image

Annotation_data/ after saving: 
image

Annotation_data/ after retraining: 
image

Evaluation metrics after retraining: 
image

Retrainer panel: 
image

# Testing

I tested it on my local environment and went through the flow of annotating objects, saving the objects, retraining the model via the button, and switching to the new model via the button. 

I verified that the resulting COCO json file has the correct annotation info on pycoco_locobot.ipynb. 
image

# Limitations

* From my own testing, once the new model is switched in and active, it **detects many more objects than normal** (100+ vs. 1-4 normally). This may be because of the tiny amount of annotations I’ve put into the training data. See images below. 
* The entire retraining process relies on files in the annotation_data/ directory, and the **paths for the annotation_data directory are hardcoded** through this PR. Ideally, we would send in the annotation_data path as an arg similarly to how Perception uses the path for agents/locobot/models/perception. 
* I included a new parameter into human_pose.py that determine which path it gets the keypoints from. The keypoints are currently downloaded into agents/locobot/models/perception/ and copying the file to the new folder is expensive, which is why **the human_pose path is currently hardcoded.** 
* **Retraining requires at least 3 annotations** since otherwise, the test set would have 0 samples. 
* There’s **little visual feedback** for the user. Once the “Save”, “Retrain”, or “Switch” buttons are pressed, there’s no immediate message saying that the backend has received the command besides in the console. 
* Annotations are saved when label prop is run. This means that if a user annotates a set of objects, then clicks "Save", the **set of objects they just annotated will not be saved**. 
  * Edit: fixed this
* Many of the **hyperparameters and settings are fixed**. For example, there is no way for the user to specify learning rate, train/test split ratio, regularization, etc. An extension can be adding a panel to the dashboard to input settings to adjust the retraining. 
  * Edit: Done! Can include more hyperparameters in the future

image
image

# Checklist:


droidlet/perception/robot/handlers/detector.py
```diff
@@ -28,8 +29,10 @@
 
 lvis_yaml = "configs/mask_rcnn_R_101_FPN_1x.yaml"
 detector_weights = "model_999.pth"
-properties = "prop.pickle"
-things = "things.pickle"
+properties_default = "prop.pickle"
```
anuragprat1k, line 32: as a follow-up to this PR, could you get rid of the defaults here? this will need updating the models tarball on s3, the protocol for which you can find here https://www.internalfb.com/intern/wiki/Droidlet/Protocol_for_uploading_models_datasets/

agents/locobot/locobot_agent.py
```diff
@@ -14,14 +14,25 @@
 import cv2
 from imantics import Mask, Polygons
 import numpy as np
+from PIL import Image
+from pathlib import Path
+import json
+from pycococreatortools import pycococreatortools
```
anuragprat1k, line 20: do we need pycococreatortools as a new dependency in requirements.txt?

anuragprat1k (changes requested):
Looks great! Thanks for the detailed PR description too.

Requesting a couple minor changes, and left some ideas for follow-up tasks here that would help.

anuragprat1k (approved):

PR #610, opened 2021-07-28
Adding a parse_chat step to core agent loop

anuragprat1k:
# Description

Issue
* In the spirit of our recent [refactor push](https://github.com/facebookresearch/droidlet/issues/337), we want the perceptual modules to be stateless. But the current [perceive() ](https://github.com/facebookresearch/droidlet/blob/main/agents/loco_mc_agent.py#L281) implementation in loco_mc_agent does not allow for a generic way for agents to pass inputs to their perceptual modules. For reference, it calls the perceptual modules as  

```
for v in self.perception_modules.values():
            v.perceive(force=force)
``` 
* Moreover, [loco_mc_agent.perceive()](https://github.com/facebookresearch/droidlet/blob/main/agents/loco_mc_agent.py#L281) also processes the incoming chat. This entanglement does not allow for a generic way to fetch inputs that `self.perception_modules` might need.

Fix
* I pulled out the chat perceive into a separate function `parse_chat`. So the core event loop has an additional step now which is separate from the rest of perceive.
* This allows downstream agents to implement their custom perceive functions to fetch states that perception might need, or use the default one if they don't need to. The locobot agent implements a custom perceive already to fetch states for its perceptual modules. [See here](https://github.com/facebookresearch/droidlet/blob/main/agents/locobot/locobot_agent.py#L220).

Fixes # (issue)

This also fixes https://github.com/facebookresearch/droidlet/issues/577 

## Type of change

- [x] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

# Testing

Tested that commands now work on the locobot agent 
Screen Shot 2021-07-28 at 7 11 04 PM

# Checklist:

### DIFF
--- a/agents/core.py
+++ b/agents/core.py
@@ -23,6 +23,7 @@ class BaseAgent:
                 self.handle_exception(e)
 
     def step(self):
+        self.parse_chat()
         self.perceive()
         self.memory.update(self)
         # maybe place tasks on the stack, based on memory/perception
@@ -30,6 +31,12 @@ class BaseAgent:
         # step topmost task on stack
         self.task_step()
         self.count += 1
+
+    def parse_chat(self):
+        """
+        Parse chat.
+        """
+        raise NotImplementedError
 
     def perceive(self):
         """
--- a/agents/loco_mc_agent.py
+++ b/agents/loco_mc_agent.py
@@ -278,7 +278,7 @@ class LocoMCAgent(BaseAgent):
         # n hundreth of seconds since agent init
         return self.memory.get_time()
 
-    def perceive(self, force=False):
+    def parse_chat(self):
         # NOTE: the processing chats block here
         # will move to chat_parser.perceive() once Soumith's changes are in
         start_time = datetime.datetime.now()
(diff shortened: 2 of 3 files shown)


agents/core.py
```diff
@@ -31,6 +32,12 @@ def step(self):
         self.task_step()
         self.count += 1
 
+    def parse_chat(self):
```
aszlam, line 35: why do we need this here?  sorry I don't understand, these are supposed to go in .perceive()

agents/loco_mc_agent.py
```diff
@@ -278,7 +278,7 @@ def get_time(self):
         # n hundreth of seconds since agent init
         return self.memory.get_time()
 
-    def perceive(self, force=False):
+    def parse_chat(self):
```
aszlam, line 281: same with this... these are supposed to go in the .perceive()

aszlam:
ok, at a high level, I strongly disagree with the approach in the PR- semantic parsing is supposed to go in the .perceive().  if the problem is that some state is needed, store whatever state is needed in memory.   to me this looks like just pulling the bits you don't like out of perceive and putting them in some other method, rather than fixing the issue...

aszlam:
look here:
https://github.com/facebookresearch/droidlet/blob/0ee6fa533318802538395203d53b295e2954382e/agents/loco_mc_agent.py#L310


in the craftassist agent were are doing this by calling super.perceive() here
https://github.com/facebookresearch/droidlet/blob/main/agents/craftassist/craftassist_agent.py#L234

aszlam:
In a few iterations, all of this will be properly handled by agent in the sense that all memory writes and reads from the parser will be mediated by the agent- for now, I think its best to keep this on locobot/robot how it is in craftassist agent

anuragprat1k:
Fixing it in https://github.com/facebookresearch/droidlet/pull/614 instead.

PR #619, opened 2021-08-02
Hollis/offline dashboard

hollisma:
# Description

- The other vision tools being built this summer (label annotating, label propagating, and retraining) can all be done with an isolated dataset. The dashboard currently requires an agent to be running, so this milestone is to let users annotate and retrain on a separate dataset not attached to an agent.
- Even though the offline dashboard does not use an agent on the backend, it still needs a backend to read/write to files and run label prop, so I created a dummy backend in ./server.py which can be run with `python server.py` .
- User flow
    - User clicks new "Offline" panel on right side, which prompts the user for a directory path. The directory needs three components: a depth folder, rgb folder, and a data.json mapping data point ids to the x, y, and yaw data. Each depth map is assumed to be in the `0000x.npy` format while rgb images are to be in the `0000x.jpg` format.
    - The first depth-image-pose instance will be displayed on the rgb, depth, and object dashboard feeds. Two buttons allow users to sequentially navigate through the instances. On any frame, users can draw annotations, which are then propagated to any other frame the user navigates to and saved to disk, assuming the new frame doesn't already have annotations.
    - Users can still save the annotations and retrain a model that can be found at the following filepath: `annotation_data/model/model_999.pth`

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

New offline panel: 
image

Offline dashboard view: 
image

Offline dashboard start: 
Offline Dashboard Start

Navigation: 
Offline Dashboard Navigation

Label propagation: 
Offline Dashboard Label Prop

# Testing

Tested locally using files from `checkpoint/apratik/ActiveVision/active_vision/replica_random_exploration_data/apartment_0`. Made sure that the above flow works. 

To test, launch the pyro habitat in the docker container, run `python server.py` in the root folder, and run `yarn start` in dashboard/web. Navigate to the Offline panel on the right and enter the filepath to the data. 

# Limitations / Improvements

- I decided to only run label prop when the new frames do not already have any annotations because otherwise, backtracking would duplicate annotations. Users might want to still propagate labels on frames that have annotations, so we can add a toggle for this.
- In order to get to the final frame, users need to navigate through the frames one by one. Can include a navigation tool to jump to a specific frame.
- We should build support for videos in the future. A nice addition to accompany this would be to automatically go through the frames so that users don't need to manually go through each frame. Could be a toggle and done at a pace that the user sets.

# Checklist:

### DIFF
--- a/agents/locobot/label_prop.py
+++ b/agents/locobot/label_prop.py
@@ -23,6 +23,21 @@ from detectron2.evaluation import COCOEvaluator, inference_on_dataset
 from droidlet.perception.robot import LabelPropagate
 
 def label_propagation(postData): 
+    """
+    postData: 
+        prevRgbImg: source rgb image
+        prevDepth: source depth object
+            depthImg: depth map image
+            depthMax: maximum value in original depth map
+            depthMin: minimum value in original depth map
+        depth: current depth object
+            same properties as prevDepth
+        objects: array of objects
+        prevBasePose: source base pose data
+        basePose: current base pose data
+
+    Returns a dictionary mapping objects ids to the objects with new masks
+    """
 
     # Decode rgb map
     rgb_bytes = base64.b64decode(postData["prevRgbImg"])
@@ -44,15 +59,8 @@ def label_propagation(postData):
     src_depth = np.array(depth_imgs[0])
     cur_depth = np.array(depth_imgs[1])
 
-    # Convert mask points to mask maps then combine them
-    src_label = np.zeros((height, width)).astype(int)
-    for n, o in enumerate(postData["objects"]): 
-        poly = Polygons(o["mask"])
-        bitmap = poly.mask(height, width)
-        for i in range(height): 
-            for j in range(width): 
-                if bitmap[i][j]: 
-                    src_label[i][j] = n + 1
+    # Labels map
+    src_label = mask_to_map(postData["objects"], height, width)
 
     # Attach base pose data
     pose = postData["prevBasePose"]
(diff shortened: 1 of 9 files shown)


droidlet/dashboard/web/src/StateManager.js
```diff
@@ -106,7 +108,7 @@ class StateManager {
     this.setTurkWorkerId(turkWorkerId);
 
     // set default url to actual ip:port
-    this.default_url = window.location.host;
+    // this.default_url = window.location.host;
```
anuragprat1k, line 111: uncomment this line?

agents/locobot/label_prop.py
```diff
@@ -0,0 +1,348 @@
...
+        save_coco(test_path, info, licenses, y_images, y_annots, categories)
+
+    # 2) Use train/test files to retrain detector
+    # Adapted from train_detector.ipynb
+    dataset_name = "annotation_coco"
+    image_dir = base_path + "rgb/"
+    train_data = dataset_name + "_train"
+    test_data = dataset_name + "_test"
+
+    if train_data in DatasetCatalog.list():
```
anuragprat1k, line 278: did `DatasetCatalog.clear()` and `MetadataCatalog.clear()` not work?

droidlet/dashboard/web/src/StateManager.js
```diff
@@ -615,6 +663,102 @@ class StateManager {
     });
   }
 
+  goOffline(filepath) {
```
anuragprat1k, line 666: can you add docstrings for each of these functions?

server.py
```diff
@@ -0,0 +1,169 @@
...
+        @sio.on("get_offline_frame")
+        def get_offline_frame(sid, data): 
+            rgb, depth = LP.get_offline_frame(data)
+            sio.emit("rgb", rgb)
+            sio.emit("depth", depth)
+
+
+if __name__ == "__main__":
+    
+    base_path = os.path.dirname(__file__)
```
anuragprat1k, line 154: how is this server kicked off while the dashboard is running?

server.py
```diff
@@ -0,0 +1,169 @@
...
+)
+
+class OfflineInstance(): 
+
+    def __init__(self): 
+        self.init_event_handlers()
+    
+    def init_event_handlers(self): 
+        @sio.on("offline_label_propagation")
+        def offline_label_propagation(sid, postData):
```
anuragprat1k, line 41: how much of this code is a duplication of the online label prop? can we reduce code duplication here?

anuragprat1k (changes requested):

agents/locobot/label_prop.py
hollisma: I put these changes in the hollis/fix_retraining branch but I can include them here as well

server.py
hollisma: Server.py is the backend for the offline dashboard and is the agent-free replacement for locobot_agent.py

server.py
hollisma: The differences are in how we read the input rgb/seg data, which account for about a third of the lines in the offline_label_propagation and offline_save_rgb_seg functions. I'll replace the duplicate code and refactor it to the label_prop file.

anuragprat1k (approved):

PR #715, opened 2021-09-20
Unit test for NLU model load and parse

kavyasrinet:
# Description

This PR does two things:
1. Adds in a unit test to make sure NLU model loads and performs inference in the current environment. This has caused some regressions lately - adding in a test.
2. Variable renaming in some files - TTAD was a very old convention.

## Type of change

- [ ] Refactor
- [x] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

# Testing

Tested locally, on devfair and CI

# Checklist:

### DIFF
--- a/droidlet/perception/semantic_parsing/tests/test_nsp_loading.py
+++ b/droidlet/perception/semantic_parsing/tests/test_nsp_loading.py
@@ -12,20 +12,26 @@ NLU_DATA_DIR = os.path.join(
 
 class TestNSPModel(unittest.TestCase):
     def setUp(self):
-        nsp_data_dir = NLU_DATA_DIR
-        nsp_model_dir = os.path.join(NLU_MODEL_DIR, "ttad_bert_updated")
-        self.model = Model(model_dir=nsp_model_dir, data_dir=nsp_data_dir)
+        self.nsp_model_dir = os.path.join(NLU_MODEL_DIR, "ttad_bert_updated")
+        self.model = Model(model_dir=self.nsp_model_dir, data_dir=NLU_DATA_DIR)
 
     def test_model_parse(self):
         chat = 'come here'
         logical_form = self.model.parse(chat=chat)
         self.assertEqual(type(logical_form), dict)
+        self.assertTrue("dialogue_type" in logical_form)
         chat = 'hello'
         logical_form = self.model.parse(chat=chat)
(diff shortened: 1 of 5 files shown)


kavyasrinet:
Oh no, just realized that I forgot to add reviewers to this :(

droidlet/perception/semantic_parsing/tests/test_nsp_loading.py
```diff
@@ -0,0 +1,31 @@
...
+class TestNSPModel(unittest.TestCase):
+    def setUp(self):
+        nsp_data_dir = NLU_DATA_DIR
+        nsp_model_dir = os.path.join(NLU_MODEL_DIR, "ttad_bert_updated")
+        self.model = Model(model_dir=nsp_model_dir, data_dir=nsp_data_dir)
+
+    def test_model_parse(self):
+        chat = 'come here'
+        logical_form = self.model.parse(chat=chat)
+        self.assertEqual(type(logical_form), dict)
```
anuragprat1k, line 22: Couple of thoughts here:
1. Is there something more than a type check we can do here? 
2. Can you also add a test for a case when model loading throws an exception - for eg when model_dir is wrong / empty etc.

anuragprat1k (changes requested):
Good idea! Requesting a tiny change.

kavyasrinet:
@anuragprat1k can you re-review please ?

droidlet/perception/semantic_parsing/tests/test_nsp_loading.py
anuragprat1k: Yea, model testing is tricky. I thought just type check might not be sufficient because it would pass even when an empty dict is returned. If an empty dict means that the NSP is broken in some way, then we should add a stronger check (since we wouldn't want this test to pass then). If not, then this is fine.

anuragprat1k (approved):

droidlet/perception/semantic_parsing/tests/test_nsp_loading.py
kavyasrinet: Yeah I added a stricter check to all three : `self.assertTrue("dialogue_type" in logical_form)` since we expect "dialogue_type" key in the dict at the minimum

PR #758, opened 2021-10-22
[Polymetis] Add Cartesian impedance control

exhaustin:
# Description

- Add working Cartesian PD controller
- Deduplicate code across different `torchcontrol.policies` modules
- Change all ee move to applications to use Cartesian impedance controllers
- Add argument type hints that were left out in previous PR

## Type of change

Please check the options that are relevant.

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Proposes a change (non-breaking change that isn't necessarily a bug)
- [x] Refactor
- [x] New feature (non-breaking change that adds a new functionality)
- [ ] Breaking change (fix or feature that would break some existing functionality downstream)
- [ ] This is a unit test
- [ ] Documentation only change
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

- [x] I want a thorough review of the implementation.
- [ ] I want a high level review. 
- [ ] I want a deep design review.

## Before and After

If this PR introduces a change or fixes a bug, please add before and after screenshots (if this is causes a UX change) or before and after examples(this could be plain text, videos etc ) to demonstrate the change.

If this PR introduces a new feature, please also add examples or screenshots demonstrating the feature.

# Testing

Please describe the tests that you ran to verify your changes and how we can reproduce.

# Checklist:

- [x] I have performed manual end-to-end testing of the feature in my environment.
- [x] I have added Docstrings and comments to the code.
- [ ] I have made changes to existing documentation where needed.
- [x] I have added tests that show that the PR is functional.
- [x] New and existing unit tests pass locally with my changes.
- [ ] I have added relevant collaborators to review the PR before merge.
- [ ] [Polymetis only] I have verified that all scripts in `tests/scripts` runs on hardware.

PR #806, opened 2021-11-11
[Polymetis] Refactor planning & Cartesian control

exhaustin:
Stacked PR on top of #758

### Previous nested module structure
```
JointSpacePD < JointImpedanceControl < JointPlanExecutor < JointSpaceMoveTo
                                                         < CartesianTargetJointMoveTo
CartesianSpacePDFast < CartesianImpedanceControl < CartesianSpaceMoveTo
```

Planners are "planning modules" within `toco.modules`

### New nested module structure
```
JointSpacePD < JointImpedanceControl
             < JointTrajectoryExecutor < JointSpaceMoveTo
                                       < CartesianTargetJointMoveTo
CartesianSpacePDFast < CartesianImpedanceControl
                     < EndEffectorTrajectoryExecutor < CartesianSpaceMoveTo

```

Planners are callable functions within `toco.planning`


1heart (approved):
If I can summarize the logic, it's just that the `.*ImpedanceControl` modules can generate the plan and save it using `toco.planning.*` functions without a `.*TrajectoryExecutor`, correct?

exhaustin:
> If I can summarize the logic, it's just that the `.*ImpedanceControl` modules can generate the plan and save it using `toco.planning.*` functions without a `.*TrajectoryExecutor`, correct?

A better summary would be:
- `.*ImpedanceControl` are policies that stabilize around a fixed state that can be updated
- `.*TrajectoryExecutor` are policies that stabilize around a given trajectory given as a list of states during initialization
- `toco.planning.*` provide helper functions that generate these trajectories from parameters

1heart:
Okay, I think the structure that makes semantic sense.

I'm not 100% sure how much simpler this is -- it seems to me that the `MoveTo` policies are still thin layers around `JointTrajectoryExecutor`. Is there another layer of indirection we can remove? Maybe `JointTrajectoryExecutor` can just take in a flag for joint or cartesian planning? Just thinking out loud, since we're taking the time to refactor.

exhaustin:
No yeah it's great to iterate on this since I agree it still seems complicated to me.

> I'm not 100% sure how much simpler this is

Currently I made the tradeoff of having some code duplication and omitting one structural layer (between `.*ImpedanceControl` and `.*TrajectoryExecutor`

> it seems to me that the `MoveTo` policies are still thin layers around `JointTrajectoryExecutor`.

Indeed, and after thinking about it I'm going to omit the `MoveTo` policies, since anyone looking to move to a specific pose would be better off using the macros in `RobotInterface` instead of instantiating the custom policies. This will make `RobotInteface` methods slightly longer.


> Is there another layer of indirection we can remove? Maybe `JointTrajectoryExecutor` can just take in a flag for joint or cartesian planning?

I would like to keep the `.*TrajectoryExecutor` layer and exclude planning from it. The reason being: in addition to joint space and cartesian space min-jerk policies, I'm working on a moveit planner integration that would output a trajectory that avoids obstacles. More joint movement plans like this could also be included in the future (such as DMPs), and thus we might have to keep adding flags to the trajectory executors, which wouldn't be sustainable.

1heart:
That sounds reasonable to me, thanks!

PR #816, opened 2021-11-16
[Polymetis] Franka gripper fix

exhaustin:
# Description

Apparently `franka::Gripper` hates it when being interrupted while executing commands (`move` or `grasp`). 
Terminating the previous command using `franka::Gripper::stop` to terminate the previous command didn't work well either.

Final fix consists of gRPC calls that are always blocking.
`GripperInterface` caches the latest gripper command and only sends it after the previous command has completed.

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

# Testing

Tested on both the Franka gripper and the Robotiq gripper

# Checklist:

- [x] [Polymetis only] I have verified that all scripts in `tests/scripts` runs on hardware.


1heart (commented):
Seems like a gripper test is broken?

In addition:

- Graceful thread shutdown, maybe during `__exit__`?
- Might want to use a size-1 queue for thread safety and simplicity, in case people are calling a lot of `_send_gripper_command`s in a row. I am not sure that this code is thread safe, e.g. a new command may occur right before line 47 and get deleted. Regardless, probably can decrease the number of `Event` flags to keep track of.
- Add a simple test

exhaustin:
> * Might want to use a size-1 queue for thread safety and simplicity, in case people are calling a lot of `_send_gripper_command`s in a row. I am not sure that this code is thread safe, e.g. a new command may occur right before line 47 and get deleted. Regardless, probably can decrease the number of `Event` flags to keep track of.

I don't see how a size-1 queue would help since managing the length of the queue would essentially be the same as replacing the variable.
Line 47 can actually be removed, since we're using `update_event` to keep track of cache updates and `done_event` to keep track of command execution (for blocking use cases). I don't see how we can remove any one of these events.

1heart:
[Python queues](https://docs.python.org/3/library/queue.html#queue.Queue) have a `maxsize` input parameter which you can set to 1 -- I think you can just get rid of all the `Event` objects, call `q.get()` in the child thread in a loop, and simply add to the queue and get thread safety for free.

exhaustin:
Also, what are the benefits of shutting down gracefully?
I think having it as a daemon thread makes sense because:
- The command execution thread doesn't hold any complicated resources that would cause problem during shutdown.
- It'll eliminate the need to track one extra "stop event".
- Daemon threads are guaranteed to not block the termination of the main thread.

1heart:
Ah okay, I missed that -- yep, daemon sounds good to me!

polymetis/polymetis/python/polymetis/gripper_interface.py
```diff
@@ -3,6 +3,8 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 import threading
+import queue
+import time
```
1heart, line 7:
```suggestion
```

1heart (approved):
Cool thanks!

PR #817, opened 2021-11-18
[Hello Robot] add look_at, fix point-cloud transforms, add 3D viewer and changes to teleop script, other bug-fixes

soumith:
# Description

As a main theme, this PR furthers along the Hello Robot Stretch integration.
It adds look_at, and fixes bugs in point-cloud transforms, base movement.

Additionally, it improves the hello_teleop script with invoke functions using the text-box.

Unrelated to the Hello Robot integration, there are a few changes;
1. In memory, save the masks for objects, not just the bounding-boxes
1. Object Deduplicator had a bug where it was updating multiple `eid`s in memory
1. The dashboard visualizer for objects has flickering of colors removed, by keying the colors on `eid`
1. In the dashboard, persons are not displayed in the Object pane (and only in the Human pane). This helps disentangle concerns (persons had bounding-boxes that occupied the entire image, weirdly)

The commits are atomic, somewhat small, and have been made to be easily reviewed.


droidlet/lowlevel/robot_mover_utils.py
```diff
@@ -110,3 +111,32 @@ def xyz_pyrobot_to_canonical_coords(xyz):
 def xyz_canonical_coords_to_pyrobot_coords(xyz):
     """converts 3D coords from canonical to pyrobot coords."""
     return xyz @ np.linalg.inv(pyrobot_to_canonical_frame)
+
+def rotation_matrix_x(a):
```
aszlam, line 115: could we have a special file for rotations/spatial transforms outside of the generic mover_utils?  eventually would want to get to some unified spatial transform interface

droidlet/memory/robot/loco_memory_nodes.py
```diff
@@ -46,9 +46,12 @@ def create(cls, memory, detected_obj) -> str:
             detected_obj.get_xyz()["z"],
             cls.NODE_TYPE,
         )
+        # TODO: remove the round-about serialization,
+        # and improve performance
+        struct = detected_obj.to_struct()
```
aszlam, line 51: so this is the place where the test is failing rn, I think bc there is a mock detection object defined in the test that doesn't have https://github.com/facebookresearch/fairo/blob/hello_robot/droidlet/perception/robot/handlers/detector.py#L193 
defined.   I wonder if we could use the "official" detection object somehow in the test instead of mocking it?

agents/locobot/hello_teleop.py
```diff
@@ -72,46 +129,80 @@ def unstop_robot(sid):
...
+            movement[0] += float(move_dist)
+            print("action: FORWARD", movement)
+            mover.move_relative([movement], blocking=False)
         elif command == "MOVE_BACKWARD":
-            movement[0] -= move
-            print("action: BACKWARD")
+            movement[0] -= float(move_dist)
+            print("action: BACKWARD", movement)
+            mover.move_relative([movement], blocking=False)
         elif command == "MOVE_LEFT":
```
anuragprat1k, line 156: is this better named 'turn left' and 'turn right'? As I understand, the semantic parser uses that semantic. @aszlam to confirm if move left is the same as turn left.

droidlet/dashboard/web/src/components/LiveObjects.js
```diff
@@ -38,8 +38,9 @@ class LiveObjects extends React.Component {
     this.onNextFrame = this.onNextFrame.bind(this);
 
     this.initialState = {
-      height: props.height,
-      width: props.width,
+      height: 640, //  props.height,
```
anuragprat1k, line 41: is this intentional? the height width will be different for our habitat setup.

droidlet/lowlevel/hello_robot/hello_robot_mover.py
```diff
@@ -56,30 +56,74 @@ def __init__(self, ip=None):
...
+        cam_transform = self.bot.get_camera_transform().value
+        cam_pos = cam_transform[0:3, 3]
+        # convert cam_pos to canonical co-ordinates
+        cam_pos = [pos[0], cam_pos[2], pos[1]]
+
+        logging.info(f"Current base state (x, z, yaw): {pos}, camera state (x, y, z): {cam_pos}")
+        logging.info(f"looking at x,y,z: {target}")
+        
+        pan_rad, tilt_rad = get_camera_angles(cam_pos, target)
+        tilt_rad = -tilt_rad
```
anuragprat1k, line 110: please add a comment for this sign inversion. why is this needed?

droidlet/lowlevel/hello_robot/remote/utils.py
```diff
@@ -94,44 +117,76 @@ def goto(robot, xyt_position=None, translation_threshold=0.1, dryrun=False, dept
...
+        # so, it will be:
+        #  0    <= angle <= pi/2   :  if +x, +y
+        # +pi/2 <= angle <= pi     :  if -x, +y
+        # -pi   <= angle <= -pi/2  :  if -x, -y
+        # -pi/2 <= angle <= 0      :  if +x, -y
+
+        # if theta_1 is between 0 and pi / 2, then you will turn left by a bit and move forward
+        # if theta_1 is between -pi/2 and 0, then you will turn right by a bit and move forward
+
         if theta_1 > pi / 2:
```
anuragprat1k, line 150: do you need a less than check here `theta_1 <= pi`?

anuragprat1k (approved):

agents/locobot/hello_teleop.py
aszlam: "move left" would be interpreted as a left "strafe", not a left turn

droidlet/lowlevel/robot_mover_utils.py
soumith: there is a file in `droidlet/lowlevel/rotation.py`. I'm moving these to that file.

droidlet/dashboard/web/src/components/LiveObjects.js
soumith: good catch, this wasn't intentional, so reverted these changes (which are not needed)

droidlet/memory/robot/loco_memory_nodes.py
soumith: i've fixed the test.
also this was a really roundabout way to serialize, which is why I added the TODO.
I've cleaned it up properly (and removed the TODO)

agents/locobot/hello_teleop.py
soumith: changing this also depends on changing locobot_agent.py.
I've made a note for myself to send in a separate patch to change all the instances of MOVE_LEFT -> TURN_LEFT

Just FYI, in the teleop, we hook up the textbox to a raw pattern-match, i.e. it doesn't invoke an interpreter or parser.

droidlet/lowlevel/hello_robot/hello_robot_mover.py
soumith: added a comment

droidlet/lowlevel/hello_robot/remote/utils.py
soumith: atan2 wont return anything greater than  `pi`, because it will then be in the quadrant `-pi <= angle <= -pi/2`

Here's an example:

```
In [25]: x, y = -1, 0; math.atan2(y, x)
Out[25]: 3.141592653589793

In [26]: x, y = -1, -0.001; math.atan2(y, x)
Out[26]: -3.1405926539231266
```

PR #825, opened 2021-11-22
De-duplicate teleop scripts between hello and habitat

soumith:
This is a simple clean-up PR.
It de-duplicates these two scripts, and deletes one of them


agents/locobot/teleop.py
```diff
@@ -182,9 +222,9 @@ def test_command(sid, commands, yaw_velocity):
             "depthMin": serialized_image["depth_min"],
         })
 
-        pcd = mover.get_current_pcd(in_global=True)
-        points = pcd[0]
-        colors = pcd[1]/255.
+
+        points, colors = rgb_depth.ptcloud.reshape(-1, 3), rgb_depth.rgb.reshape(-1, 3)
+        colors = colors / 255.
```
anuragprat1k, line 227: is this normalization necessary here? Was thinking when `reshape` creates a new view versus a copy, and changing the values (by normalization) would create a new copy, right? Do we need that extra overhead?

agents/locobot/teleop.py
```diff
@@ -231,78 +270,83 @@ def test_command(sid, commands, yaw_velocity):
...
         robot_orientation.compute_vertex_normals()
-        robot_orientation.paint_uniform_color([0.1, 0.9, 0.1])
+        robot_orientation.paint_uniform_color([1.0, 0.5, 0.1])
         
-        robot_orientation.translate([y, -x, 0.5], relative=False)
+        robot_orientation.translate([y, -x, 0.], relative=False)
         robot_orientation.rotate(o3d.geometry.get_rotation_matrix_from_axis_angle([0, math.pi/2, 0]))
-        robot_orientation.rotate(o3d.geometry.get_rotation_matrix_from_axis_angle([0, 0, yaw]))        
+        if yaw != 0:
+            robot_orientation.rotate(o3d.geometry.get_rotation_matrix_from_axis_angle([0, 0, yaw]))
```
anuragprat1k, line 278: what are the two rotations for here? also, do we need the if check for the second rotation?

agents/locobot/teleop.py
```diff
@@ -231,78 +270,83 @@ def test_command(sid, commands, yaw_velocity):
...
-        # obstacles = np.asarray(obstacles)
-        # obstacles = np.concatenate((-obstacles[:, [1]], -obstacles[:, [0]], np.zeros((obstacles.shape[0], 1))), axis=1)
-        # obspcd = o3d.geometry.PointCloud()
-        # obspcd.points = o3d.utility.Vector3dVector(obstacles)
-        # obspcd.paint_uniform_color([1.0, 0., 0.])
-        # obsvox = o3d.geometry.VoxelGrid.create_from_point_cloud(obspcd, 0.03)
-        # o3dviz.put('obstacles', cmd, obsvox)                
-            
+            # get the SLAM goals
+            goal_loc, stg = None, None # mover.bot.get_slam_goal()
```
anuragprat1k, line 300: hook up the explore to a command as well (like 'explore')? this is a no-op right now as I understand and needs to be uncommented to work, right?

anuragprat1k (approved):
Looks good! Feel free to resolve the comments as you see fit.

agents/locobot/teleop.py
soumith: reshape is like a `maybe_copy`. It creates a copy only when a view is not possible.
Here, it's creating views, so colors = colors / 255. is needed

agents/locobot/teleop.py
soumith: will add comments, but this is the cylinder representating the orientation of the robot, so it needs to be first made parallel to the floor, and then rotated by `yaw`

agents/locobot/teleop.py
soumith: will make this work again and uncomment in a subsequent PR

PR #829, opened 2021-11-22
update split masks to separate old and new data

ethan-carlson:
# Description

Changes the way NSP retraining split masks are generated so that old and new data are handled separately.

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

- [X] I want a thorough review of the implementation.

## Before and After

Previously there were two options for generating the split masks:
 - Resampled - old+new data were mixed and split into new train/valid/test sets
 - Not resampled - old training data and all new data are used for training, but the valid and test sets are only the old versions

This creates the potential issue that a model trained with --resampled=false could see data used in training in evaluation if evaluated using the other masks.

With this change, new 80/10/10 train/valid/test masks are generated for the new data, and the user can specify which sets to mix or not.  Any of the three (train, valid and test) may contain either only old data, only new data, or a mix of the two.  Regardless of the choices, the splits remain constant so that data used in training is never seen in in valid or test.

# Testing

Run nsp_retrain_infra.py from the command line to verify that splits are generated as expected.

# Checklist:

- [X] I have added Docstrings and comments to the code.
- [X] New and existing unit tests pass locally with my changes.

### DIFF
--- a/droidlet/tools/hitl/nsp_retrain_infra.py
+++ b/droidlet/tools/hitl/nsp_retrain_infra.py
@@ -7,13 +7,11 @@ import logging
 import os
 import shutil
 import time
-from subprocess import Popen, PIPE, TimeoutExpired
+from subprocess import Popen, PIPE
 import boto3
 from datetime import datetime
 import torch
 import random
-
-from typing import List
 
 from droidlet.tools.hitl.data_generator import DataGenerator
 from droidlet.tools.hitl.job_listener import JobListener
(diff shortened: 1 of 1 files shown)


snyxan:
maybe we should also save the masks (just for convenience since we can always construct the same ones using the same random seed) and upload it to S3.

snyxan:
also we need a script to just evaluate existing models with different test dataset.@aszlam would you able to help Ethan with that here?

droidlet/tools/hitl/nsp_retrain_infra.py
```diff
@@ -161,28 +161,39 @@ def create_mask(self, opts, batch_id, data_filepath):
...
-            mask_designation.extend(['train'] * (total_rows - len(mask_designation)))
-            random.shuffle(mask_designation)
-            train_mask = [True if x == 'train' else False for x in mask_designation]
-            valid_mask = [True if x == 'valid' else False for x in mask_designation]
-            test_mask = [True if x == 'test' else False for x in mask_designation]
+
+        # Old masks are static, just extend to be the length of the new dataset
+        old_mask_filepath = os.path.join(opts.sweep_config_folder, 'split_masks.pth')
+        old_masks = torch.load(old_mask_filepath)['annotated']
+        old_train_mask = [x for x in old_masks['train']]
```
kavyasrinet, line 168: isn't this the same as : `old_train_mask = old_masks['train']` why iterate over and build a list ? 
Same for valid and test

droidlet/tools/hitl/nsp_retrain_infra.py
```diff
@@ -161,28 +161,39 @@ def create_mask(self, opts, batch_id, data_filepath):
...
+        old_train_mask.extend([False] * (total_rows - len(old_train_mask)))
+        old_valid_mask = [x for x in old_masks['valid']]
+        old_valid_mask.extend([False] * (total_rows - len(old_valid_mask)))
+        old_test_mask = [x for x in old_masks['test']]
+        old_test_mask.extend([False] * (total_rows - len(old_test_mask)))
+
+        # Set random seed for reproducability and generate masks for new data
```
kavyasrinet, line 175: might be worth adding a comment about 80, 10, 10 here for better readability

droidlet/tools/hitl/nsp_retrain_infra.py
```diff
@@ -161,28 +161,39 @@ def create_mask(self, opts, batch_id, data_filepath):
...
+        random.seed(batch_id)
+        new_mask_designation = ['test'] * int(new_data_rows * 0.1)
+        new_mask_designation.extend(['valid'] * int(new_data_rows * 0.1))
+        new_mask_designation.extend(['train'] * (new_data_rows - len(new_mask_designation)))
+        random.shuffle(new_mask_designation)
```
kavyasrinet, line 180: can we just write a function for this instead of redundant steps here ?
or something less redundant like:
```
all_masks = {}
for mask_type in ['test', 'train', 'valid']:
    all_masks[mask_type] = [False] * len(new_masks)
    for j in [i for i in range(len(new_masks)) if new_masks[i] == mask_type]:
        all_masks[mask_type][j] = True
```

droidlet/tools/hitl/nsp_retrain_infra.py
```diff
@@ -161,28 +161,39 @@ def create_mask(self, opts, batch_id, data_filepath):
...
+        new_masks = [False] * (total_rows - new_data_rows)  # Prepend False for old data
+        new_masks.extend(new_mask_designation)
+        new_train_mask = [True if x == 'train' else False for x in new_masks]
+        new_valid_mask = [True if x == 'valid' else False for x in new_masks]
+        new_test_mask = [True if x == 'test' else False for x in new_masks]
+
+        # Mix and match old and new data based on user input
+        if opts.retrain_data_splits[0] == 0: train_mask = old_train_mask
```
kavyasrinet, line 188: same as above, we should try to nuke as much redundancy in code as possible.

kavyasrinet (changes requested):

droidlet/tools/hitl/nsp_retrain_infra.py
ethan-carlson: To convert from a torch tensor to a list, but there's also a built in function to do it.

droidlet/tools/hitl/nsp_retrain_infra.py
ethan-carlson: added

droidlet/tools/hitl/nsp_retrain_infra.py
ethan-carlson: For sure, let me know if you like the way I rebuilt things in the latest commit.

droidlet/tools/hitl/nsp_retrain_infra.py
ethan-carlson: also addressed in the latest commit

kavyasrinet (approved):
Thanks for the quick fixes.

snyxan (approved):

PR #848, opened 2021-12-01
[HiTL] Updated message pane chat interface and "stoplight" HIT quality indicator

ethan-carlson:
# Description

This PR includes two updates to the dashboard, especially has it relates to Craftassist interaciton HITs:

- The first is to do away with the 'History' pane and update the 'Message' pane to have a more conventional chat interface, like you'd see eg. on your phone.  This better matches user expectations and reduces cognitive load.
- The second change introduces a "stoplight" quality indicator on the 'TurkInfo' pane shown during MTurk Craftassist interaction HITs.  The stoplight is an easy-to-understand visual display of the quality of the interaction, defined by the quantity, diversity, and creativity of the commands issued, all weighted equally.
   - The quantity score is simply the number of commands issued, with a max contribution value of 10
   - The diversity score is a measure of how different the commands are from other commands the user has issued within the same HIT.  The score is a value out of 10 based on the average minimum edit distances between the commands, not including stop words.
   - The creativity score is a measure of how different the commands are from the corpus of commands issued by other turkers previously (the static "old" dataset hosted on S3).  Similarly this is out of 10 based on average minimum edit distances between commands, not including stop words.  There is an additional penalty applied for commands that are particularly short or particularly long.
   - The stoplight score is then the sum of the logs of each of the constituent scores, scaled to be out of 10.
   - The stoplight is red for scores less than 5, yellow for scores less than 6.5, otherwise green.  These are fairly generous cutoffs in practice but should serve to disincentivize bad behavior.

The formulae and cutoffs are determined empirically to land in the right range and scale in ways that are desirable, but please offer commentary on the method and implementation.

The other final minor addition here is a text overlay on VoxelWorld before the user clicks in for the first time.  When the VoxelWorld iframe is loaded for the first time it appears grey with no content, unlike the images shown in the instructions.  The user might think it's still loading or has crashed, so a prompt to click in helps direct the user to move forward with the task.

## Type of change

- [ ] Refactor
- [X] New feature (non-breaking change that adds a new functionality)
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

- [X] I want a thorough review of the implementation.
- [X] I want a deep design review.

## Before and After

Previously the chat interface was composed of the message window and history pane together, which was kind of confusing.  These functions are now combined into the new chat interface.

The stoplight is a new feature entirely.

Photos showing the changes before and after can be seen [here.](https://docs.google.com/document/d/1JB6igdT6mMuCfRDvocmwp0Tt5WVkIuvzx1oeFCdpkkE/edit?pli=1#heading=h.pikrdtt8gd97)

# Testing

To test these features you can run the dashboard locally with the /turk.js suffix.  There is also a demo of this dashboard version running at this link (currently slightly out of date as of the commits made 12/3): http://dashboard-stoplight-0.craftassist.io/turk.html

# Checklist:

- [X] I have added Docstrings and comments to the code.
- [X] New and existing unit tests pass locally with my changes.


kavyasrinet (approved):
Thanks for adding this in Ethan! LGTM

PR #854, opened 2021-12-07
Adding a TrajectorySaverTask

anuragprat1k:
Adds a new task TrajectorySaverTask - all subclasses derived from this can save the robot trajectory, RGB, depth, and in simulation, segmentation masks.

Stack from [ghstack](https://github.com/ezyang/ghstack):
* #911
* #909
* #900
* #861
* #860
* #859
* #858
* #857
* #856
* #855
* __->__ #854

PR #855, opened 2021-12-07
get_step_target_for_straightline_move util

anuragprat1k:
Adds a util that generates a straightline path from the robot pose to the specified target.

* #911
* #909
* #900
* #861
* #860
* #859
* #858
* #857
* #856
* __->__ #855
* #854

### DIFF
--- a/droidlet/lowlevel/robot_mover_utils.py
+++ b/droidlet/lowlevel/robot_mover_utils.py
@@ -96,6 +96,35 @@ def get_move_target_for_point(base_pos, target, eps=0.5):
 
     return [targetx, targetz, yaw]
 
+def get_step_target_for_straightline_move(base_pos, target, step_size=0.1):
+    """
+    Heuristic to get step target of step_size for going to from base_pos to target
+    in a straight line. 
+    Args:
+        base_pos ([x,z,yaw]): robot base in canonical coords
+        target ([x,y,z]): point target in canonical coords
+    
+    Returns:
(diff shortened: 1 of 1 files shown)


droidlet/lowlevel/robot_mover_utils.py
```diff
@@ -96,6 +96,35 @@ def get_move_target_for_point(base_pos, target, eps=0.5):
 
     return [targetx, targetz, yaw]
 
+def get_step_target_for_straightline_move(base_pos, target, step_size=0.1):
+    """
+    Heuristic to get step target of step_size for going to from base_pos to target
+    in a straight line. 
+    Args:
+        base_pos ([x,z,yaw]): robot base in canonical coords
+        target ([x,y,z]): point target in canonical coords
```
aszlam, line 105: [x, z, yaw] is a different format than [x, y, z], no? we should name these formats differently instead of calling both "canonical coords".

droidlet/lowlevel/robot_mover_utils.py
```diff
@@ -96,6 +96,35 @@ def get_move_target_for_point(base_pos, target, eps=0.5):
...
+    
+    Returns:
+        move_target ([x,z,yaw]): robot base move target in canonical coords 
+    """
+
+    dx = target[0] - base_pos[0]
```
aszlam, line 111: make a comment here that this needs to fixed in the future with some generic rotations.py like interface

anuragprat1k:
stack merged in #910

PR #856, opened 2021-12-07
visualization function for Examine tasks

anuragprat1k:
Adds a visualization function for Examine tasks, visualizes canonical coordinates (under the env variable `VISUALIZE_EXAMINE`)

Stack from [ghstack](https://github.com/ezyang/ghstack):
* #911
* #909
* #900
* #861
* #860
* #859
* #858
* #857
* __->__ #856
* #855
* #854

PR #857, opened 2021-12-07
setting exploration goal in agent, making Explore a TrajectorySaverTask

anuragprat1k:
* Moves setting the goal for exploration to the agent. This was previously hard-coded in the navigation service.
* Adds an environment variable `CONTINUOUS_EXPLORE` to continuously keep exploring. This is useful for collecting multiple exploration trajectories.
* Changes `Explore` into a `TrajectorySaverTask`

* #911
* #909
* #900
* #861
* #860
* #859
* #858
* __->__ #857
* #856
* #855
* #854

### DIFF
--- a/droidlet/interpreter/robot/default_behaviors.py
+++ b/droidlet/interpreter/robot/default_behaviors.py
@@ -3,8 +3,47 @@ Copyright (c) Facebook, Inc. and its affiliates.
 """
 import logging
 from droidlet.interpreter.robot import tasks
+import os
+import random
+import numpy as np
 
+random.seed(2021) # fixing a random seed to fix default exploration goal
+first_exploration_done = False
+
+def get_distant_goal(x, y, t, l1_thresh=35):
+    # Get a distant goal for the slam exploration
+    # Pick a random quadrant, get 
+    while True:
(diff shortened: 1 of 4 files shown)


droidlet/interpreter/robot/default_behaviors.py
```diff
@@ -3,8 +3,47 @@
...
+    # Pick a random quadrant, get 
+    while True:
+        xt = random.randint(-19, 19)
+        yt = random.randint(-19, 19)
+        d = np.linalg.norm(np.asarray([x,y]) - np.asarray([xt,yt]), ord=1)
+        if d > l1_thresh:
+            return (xt, yt, 0)
+
+def start_explore(agent, goal):
+    global first_exploration_done
```
aszlam, line 24: why not store this in memory, or in agent?  why a global?

anuragprat1k: Agree, but wasn't sure of the best way to just have generic state (like this) in memory. Could you point me to a similar example? Or what would be a good way to do this?

anuragprat1k:
stack merged in #910

PR #858, opened 2021-12-07
reset explore changes

anuragprat1k:
Changes to reset exploration - we want to reset the map builder and reset the `_is_done_exploring` flag in slam.

Stack from [ghstack](https://github.com/ezyang/ghstack):
* #911
* #909
* #900
* #861
* #860
* #859
* __->__ #858
* #857
* #856
* #855
* #854

PR #859, opened 2021-12-07
CuriousExplore + ExamineDetectionStraightline

anuragprat1k:
# Description

* Implements `CuriousExplore` and `ExamineDetectionStraightline` as a subclass of `TrajectorySaverTask`.

The following visualization shows all the straightline explorations. Note that it does not visualizes the robot poses where the robot was exploring randomly.

straightline

Follow-ups
* The static class ExaminedMap needs to be in memory. Will spec this out.
* Add a habitat test.

* #911
* #909
* #900
* #861
* #860
* __->__ #859
* #858
* #857
* #856
* #855
* #854

### DIFF
--- a/droidlet/interpreter/robot/tasks.py
+++ b/droidlet/interpreter/robot/tasks.py
@@ -5,7 +5,9 @@ Copyright (c) Facebook, Inc. and its affiliates.
 import time
 import logging
 import numpy as np
-
+import os
+import math
+from droidlet.memory.robot.loco_memory import DetectedObjectNode
 from droidlet.task.task import Task, BaseMovementTask
 from droidlet.memory.memory_nodes import TaskNode
 
@@ -16,6 +18,10 @@ from droidlet.lowlevel.robot_mover_utils import (
     ARM_HEIGHT,
     get_camera_angles,
     TrajectoryDataSaver,
+    visualize_examine,
+    get_step_target_for_straightline_move,
+    ExaminedMap,
+    CAMERA_HEIGHT,
 )
 
 # FIXME store dances, etc.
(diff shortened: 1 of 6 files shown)


droidlet/interpreter/robot/tasks.py
```diff
@@ -5,7 +5,9 @@
 import time
 import logging
 import numpy as np
-
+import os
+import math
+from droidlet.memory.robot.loco_memory import DetectedObjectNode
```
aszlam, line 10: this isn't in loco_memory_nodes.py ?  when did that change happen?  has this been tested?

droidlet/interpreter/robot/tasks.py
```diff
@@ -359,6 +365,159 @@ def step(self):
...
         return "<TrajectorySaverTask {}>".format(self.target)
 
+class CuriousExplore(TrajectorySaverTask):
+    """use slam to explore environemt, but also examine detections"""
+
+    def __init__(self, agent, task_data):
+        super().__init__(agent, task_data)
+        self.steps = ["not_started"] * 2
+        self.task_data = task_data
+        self.goal = task_data.get("goal", (19,19,0))
```
aszlam, line 375: why (19,19, 0) ?

droidlet/interpreter/robot/tasks.py
```diff
@@ -359,6 +365,159 @@ def step(self):
...
+
+    @Task.step_wrapper
+    def step(self):
+        super().step()
+        self.interrupted = False
+        self.finished = False
+        
+        if self.steps[0] == "not_started":
+            if self.agent.mover.nav.is_done_exploring().value:
+                # clear memory
```
aszlam, line 404: can we have a separate piece of task_data that controls this behavior?  why should an explore always clear memory at its init?

droidlet/interpreter/robot/tasks.py
```diff
@@ -359,6 +365,159 @@ def step(self):
...
+                    if ExaminedMap.can_examine(x):
+                        # check for line of sight and if within a certain distance
+                        yaw_rad, _ = get_camera_angles([base_pos[0], CAMERA_HEIGHT, base_pos[1]], x['xyz'])
+                        # self.logger.info(f"{x['label']}, yaw_rad {yaw_rad}, base_pos {base_pos}")
+                        dist = np.linalg.norm(base_pos[:2]-[x['xyz'][0], x['xyz'][2]])
+                        if abs(yaw_rad - base_pos[2]) <= math.pi/4 and dist <= 3:
+                            self.logger.info(f"Exploring eid {x['eid']}, {x['label']} next, dist {dist}")
+                            return x
+                return None
+            target = pick_random_in_sight(objects, pos)
```
aszlam, line 436: should we have some logic to explore all detected objects, or otherwise encourage not looking at the same things over and over again?

droidlet/lowlevel/robot_mover_utils.py
```diff
@@ -242,10 +242,72 @@ def visualize_examine(agent, robot_poses, object_xyz, label, obstacle_map, pts):
...
+        plt.plot(pts[:,0], pts[:,1], 'y--')
     
     # TODO: visualize robot heading 
     
     plt.savefig("{:04d}.jpg".format(vis_count))
-    vis_count += 1
\ No newline at end of file
+    vis_count += 1
+
+class ExaminedMap:
```
aszlam, line 254: I am ok with leaving this like this for the PR; but I am going to move to memory immediately after this is merged

droidlet/memory/robot/loco_memory.py
```diff
@@ -94,3 +94,8 @@ def get_detected_objects_tagged(self, *tags) -> List["DetectedObjectNode"]:
     def add_dance(self, dance_fn, name=None, tags=[]):
         # a dance is movement determined as a sequence of steps, rather than by its destination
         return DanceNode.create(self, dance_fn, name=name, tags=tags)
+
+    def clear(self, objects):
```
aszlam, line 98: would be better to just overload .forget I think

droidlet/interpreter/robot/tasks.py
anuragprat1k: Good point. Turns out loco_memory imports all loco_memory_nodes (https://github.com/facebookresearch/fairo/blob/main/droidlet/memory/robot/loco_memory.py#L11) which is why this works. Changing this to specifically import from nodes.

droidlet/interpreter/robot/tasks.py
anuragprat1k: this was just the previous default, but getting rid of this so that we always have to ensure setting the goal in default_behaviors.py

droidlet/interpreter/robot/tasks.py
anuragprat1k: Moved the clearing memory out of the task to happen before the task is initialized (https://github.com/facebookresearch/fairo/pull/900/files#diff-4ca1ae20f4650f0a47b5e77eec5f8ae8c58be252584584165d5d82bbb3cbe49bR56). Note that this is in a later commit in the stack so you won't see the change here.

droidlet/interpreter/robot/tasks.py
anuragprat1k: We do check to ensure that the same region isn't explored more than twice (https://github.com/facebookresearch/fairo/pull/859/files#diff-59137b4e93337879456a9581c814274ac572e6abba78c8f0bbb56b07deb2a12aR260). As for exploring all detected objects, it's something we can try as a follow-up.

droidlet/memory/robot/loco_memory.py
anuragprat1k: .forget takes one memid where as this takes a list of them. Hence didn't / couldn't overload it.

droidlet/lowlevel/robot_mover_utils.py
anuragprat1k: Sounds good. Added an issue here https://github.com/facebookresearch/fairo/issues/907

anuragprat1k:
stack merged in #910

PR #860, opened 2021-12-07
util to generate circular examine path

anuragprat1k:
Adds a util to generate a circular path around a given target. It returns the path where the first point is the closest point on the circle to the robot.

Stack from [ghstack](https://github.com/ezyang/ghstack):
* #911
* #909
* #900
* #861
* __->__ #860
* #859
* #858
* #857
* #856
* #855
* #854

PR #861, opened 2021-12-07
Circular examine detection

anuragprat1k:
This commit adds the Circular Examination task. The yellow circle is the circular path generated for the robot to travel along, and the red marks the part of the circle the robot actually travels. Note again that the visualization does not have the robot poses where the robot was randomly exploring.

circle

This is a sample of the examine of the first "chair" from the gif above
chair

Follow-up:
1. There's a couple of things we can tune here to increase trajectory length - how many points are generated on the circle, and also saving the trajectory where the robot moves to the edge of the circle.
2. Habitat tests.

* #911
* #909
* #900
* __->__ #861
* #860
* #859
* #858
* #857
* #856
* #855
* #854

### DIFF
--- a/droidlet/interpreter/robot/tasks.py
+++ b/droidlet/interpreter/robot/tasks.py
@@ -22,6 +22,7 @@ from droidlet.lowlevel.robot_mover_utils import (
     get_step_target_for_straightline_move,
     ExaminedMap,
     CAMERA_HEIGHT,
+    get_circular_path,
 )
 
 # FIXME store dances, etc.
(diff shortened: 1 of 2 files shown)


droidlet/interpreter/robot/tasks.py
```diff
@@ -519,6 +520,71 @@ def step(self):
     def __repr__(self):
         return "<ExamineDetectionStraightline {}>".format(self.target['label'])
 
+
+class ExamineDetectionCircle(TrajectorySaverTask):
```
aszlam, line 524: fine for now, but I would really rather this and the straight line one to be a single class with the path type in the task data

anuragprat1k: Makes sense. Created an issue that I'll come back to in a hack session (https://github.com/facebookresearch/fairo/issues/906). Don't want to keep adding to this PR since it is becoming quite large.

anuragprat1k:
stack merged in #910

PR #866, opened 2021-12-09
Simplifying label prop

anuragprat1k:
Propagated labels are the reprojected points in the new frame. This works around ~400 times faster per propagation and is comparable in accuracy. Can improve this further using depth filtering to fill in holes, but postponing that for later. For the scope of our work this half, this should suffice. 

Below is a table that compares the accuracy and time for the tests added in this PR for noisy and non-noisy data (collected from noisy, non-noisy trajectories respectively) 

|Non-noisy|Noisy|
|---|---|
|Screen Shot 2021-12-09 at 6 34 43 PM|Screen Shot 2021-12-09 at 6 38 44 PM|

Below are some before and after propagations.

| Before | After |
| --- | --- |
| out (7) | out (6) |
| out (3) | out (4) |
|  hab1_old | hab1_new |

* #867
* __->__ #866

### DIFF
--- a/droidlet/perception/robot/handlers/label_propagate.py
+++ b/droidlet/perception/robot/handlers/label_propagate.py
@@ -2,21 +2,57 @@
 Copyright (c) Facebook, Inc. and its affiliates.
 """
 from .core import AbstractHandler
-from tokenize import String
 import numpy as np
 import os
-import sys
-if "/opt/ros/kinetic/lib/python2.7/dist-packages" in sys.path:
-    sys.path.remove("/opt/ros/kinetic/lib/python2.7/dist-packages")
 import cv2
 import json
-from copy import deepcopy as copy
-from IPython import embed
-import time
-from scipy.spatial.transform import Rotation
 import glob
-import argparse
 from droidlet.lowlevel.robot_mover_utils import transform_pose
+from numba import njit
+from math import ceil, floor
+
+def compute_uvone(height, width):
+    intrinsic_mat = np.array([[256, 0, 256], [0, 256, 256], [0, 0, 1]])
+    rot = np.array([[0.0, 0.0, 1.0], [-1.0, 0.0, 0.0], [0.0, -1.0, 0.0]])
+    trans = np.array([0, 0, 0.6])
+    intrinsic_mat_inv = np.linalg.inv(intrinsic_mat)
+    img_resolution = (height, width)
+    img_pixs = np.mgrid[0 : img_resolution[0] : 1, 0 : img_resolution[1] : 1]
+    img_pixs = img_pixs.reshape(2, -1)
+    img_pixs[[0, 1], :] = img_pixs[[1, 0], :]
+    uv_one = np.concatenate((img_pixs, np.ones((1, img_pixs.shape[1]))))
+    uv_one_in_cam = np.dot(intrinsic_mat_inv, uv_one)
+    return uv_one_in_cam, intrinsic_mat, rot, trans
+
+def convert_depth_to_pcd(depth, pose, uv_one_in_cam, rot, trans):
+    # point cloud in camera frame
+    depth = (depth.astype(np.float32) / 1000.0).reshape(-1)
+    pts_in_cam = np.multiply(uv_one_in_cam, depth)
+    pts_in_cam = np.concatenate((pts_in_cam, np.ones((1, pts_in_cam.shape[1]))), axis=0)
+    # point cloud in robot base frame
+    pts_in_base = pts_in_cam[:3, :].T
+    pts_in_base = np.dot(pts_in_base, rot.T)
+    pts_in_base = pts_in_base + trans.reshape(-1)
+    # point cloud in world frame (pyrobot)
+    pts_in_world = transform_pose(pts_in_base, pose)
+    return pts_in_world
+
+@njit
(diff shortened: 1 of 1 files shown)


soumith:
i think it's 400x faster, not 40. collapsing the function out of the njit loop probably gave another 10x boost (from 0.5 seconds to 50ms)

anuragprat1k:
> i think it's 400x faster, not 40. collapsing the function out of the njit loop probably gave another 10x boost (from 0.5 seconds to 50ms)

Lol, yes. Only so much math I can do in a day I suppose.

aszlam (commented):
Could you add some comments to these?  describe the algorithms in words

droidlet/perception/robot/handlers/label_propagate.py
```diff
@@ -2,21 +2,57 @@
...
-import time
-from scipy.spatial.transform import Rotation
 import glob
-import argparse
 from droidlet.lowlevel.robot_mover_utils import transform_pose
+from numba import njit
+from math import ceil, floor
+
+def compute_uvone(height, width):
+    intrinsic_mat = np.array([[256, 0, 256], [0, 256, 256], [0, 0, 1]])
```
aszlam, line 15: nit: can these be constants?  also, can they be formatted ?  e.g. outside of the function, do

INTRINSIC_MAT = np.array([[256, 0,      256 ], 
                                                 [0,     256,  256 ], 
                                                 [0,     0,       1      ]])
                                                 
     and say what each of these constants is with a comment

aszlam: lol github code blocks do not preserve the formatting.  left edge of numbers should be aligned

droidlet/perception/robot/handlers/label_propagate.py
```diff
@@ -2,21 +2,57 @@
...
+    rot = np.array([[0.0, 0.0, 1.0], [-1.0, 0.0, 0.0], [0.0, -1.0, 0.0]])
+    trans = np.array([0, 0, 0.6])
```
aszlam, line 17: especially this one.  why .6?  the .6 should be a named constant itself

droidlet/perception/robot/handlers/label_propagate.py
```diff
@@ -2,21 +2,57 @@
...
+    pts_in_base = pts_in_cam[:3, :].T
+    pts_in_base = np.dot(pts_in_base, rot.T)
+    pts_in_base = pts_in_base + trans.reshape(-1)
+    # point cloud in world frame (pyrobot)
+    pts_in_world = transform_pose(pts_in_base, pose)
+    return pts_in_world
+
+@njit
+def get_annot(height, width, pts_in_cur_img, src_label):
+    annot_img = np.zeros((height, width))
```
aszlam, line 42: comment this fn

droidlet/perception/robot/handlers/label_propagate.py
```diff
@@ -2,21 +2,57 @@
...
+    for x in range(len(pts_in_cur_img)):
```
aszlam, line 43: is it necessary to use x as loop var _and_ as coordinate?

droidlet/perception/robot/handlers/label_propagate.py
```diff
@@ -2,21 +2,57 @@
...
+        r = int(x/width)
+        c = int(x - r*width)
+        x, y, _ = pts_in_cur_img[x]
```
aszlam, line 46: is it necessary to use x as loop var and as coordinate?

anuragprat1k: No, good catch.

droidlet/perception/robot/handlers/label_propagate.py
anuragprat1k: These are just hard-coded intrinsic and extrinsic parameter values. Eventually, I want to move all the transformation parameters and their associated utilities to a consolidated place (across Hello and Locobot). Waiting to see @soumith's changes to figure out what the best place to do that would be. Will add a todo for now.

anuragprat1k:
> Could you add some comments to these? describe the algorithms in words

Added them here 
https://github.com/facebookresearch/fairo/pull/866/files#diff-885998715b34fadd02d121d3207847293380ac21f8e010caf88282d3b151a413R72

droidlet/perception/robot/handlers/label_propagate.py
aszlam: Right, and because they are hard-coded consts, it imo is nicer that they appear at the header of the file, denoted as consts, with a short comment about what the const is.   (agree with having a consolidated file long term).

anuragprat1k: Yep, definitely will do that.

aszlam (approved):

droidlet/perception/robot/handlers/label_propagate.py
anuragprat1k: updated to constants. also self-assigned a task to streamline how we process the camera parameters https://github.com/facebookresearch/fairo/issues/901

droidlet/perception/robot/handlers/label_propagate.py
anuragprat1k: resolved in the comment below.

PR #869, opened 2021-12-10
using gitpath to access repository root

anuragprat1k:
* (to be filled)

### DIFF
--- a/agents/locobot/requirements.txt
+++ b/agents/locobot/requirements.txt
@@ -23,3 +23,4 @@ wordcloud
 fuzzywuzzy
 open3d
 numba
+git+https://github.com/maxnoe/python-gitpath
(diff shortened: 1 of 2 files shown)


kavyasrinet (approved):
Thanks for adding this in

soumith (changes requested):
adding a dependency for this is a bit of an overkill, especially a dependency that doesn't have a `pip` package either.

Add code to `droidlet/__init__.py` that defines `REPO_ROOT = os.path.abspath(os.path.dirname(__file__))` and then do `from droidlet import REPO_ROOT`

anuragprat1k:
> REPO_ROOT = os.path.abspath(os.path.dirname(__file__))

This is much neater. Done, thanks.

PR #887, opened 2021-12-13
candidate selection heuristic for active vision + tests

anuragprat1k:
1. This PR adds a candidate selection heuristic for the active vision pipeline. **For this heuristic, we are simulating how a helpful turker would pick frames to be labeled that maximize the propagation of labels we care about. More specifically, we want frames where the objects concerned are in view and relatively close.** 
2. Since this selection heuristic will be called multiple times for the same trajectory in the active vision training pipeline, the heuristic also uses a json cache to save the good candidate frames for each trajectory (this is the most expensive operation). **The caching helps reduce the call time of subsequent calls to 0.5 seconds from the initial time of 150 seconds.**

Some examples of good and bad candidates (for labels `['chair', 'cushion', 'door', 'indoor-plant', 'sofa', 'table']`)

Screen Shot 2021-12-13 at 2 18 34 PM

Screen Shot 2021-12-13 at 2 18 48 PM

* __->__ #887


examples_and_tutorials/notebooks/active_vision/candidate_selection.ipynb
```diff
@@ -0,0 +1,138 @@
+{
```
aszlam, line 1: what does this notebook do?

anuragprat1k: deleted it and added tests as discussed.

anuragprat1k:
closing this, opened an equivalent PR at https://github.com/facebookresearch/fairo/pull/905

PR #895, opened 2021-12-15
Obstacle detection for Habitat and Hello Robot

soumith:
# Description

This is a fairly large PR, but the tl;dr is that it does a few main things (and a bunch of cleanup of dead code across board).

The main stuff:

1. It adds obstacle detection based on ground-plane detection. It detects the points on the ground plane, and subsequently removes those points, and looks at the sum of total remaining points within a bounding-box in front of the robot (and if they cross a threshold, says that they are obstacles).

Screen Shot 2021-12-15 at 5 18 05 AM

Screen Shot 2021-12-15 at 5 14 51 AM

2. It hooks in the obstacle detection into the Habitat navigation, deep in the `_act(...)` function (this is similar to the Hello Robot integration's `goto` function), where instead of actually colliding with objects and then backtracking, the robot now simply detects objects in front, and then starts backtracking (the map gets updated anyways via the slam_service / mapping_service, so the robot wont take that path subsequently).

The main limitation, that we need to address in subsequent PRs is:
- The robot needs to be able to see right in front of itself, so it's camera has to be tilted down (roughly 1.5 radians for the locobot).

In a subsequent PR, one has to do add options to:
1. disable the collision-check: Unsafe Move
2. do the collision check at the beginning of moving forward, and once the immediate ground is clear, do an unsafe move forward (assume that its a static scene)
3. look closer to the robot while moving, always, but once in  a while, look up and map the world by rotating the camera all around


agents/locobot/saved_trajectories.py
```diff
@@ -0,0 +1,116 @@
...
+import json
+import math
+import cv2
+import numpy as np
+import open3d as o3d
+from droidlet import dashboard
+from droidlet.dashboard.o3dviz import o3dviz
+from droidlet.lowlevel.hello_robot.hello_robot_mover import HelloRobotMover
+
+if __name__ == "__main__":
```
aszlam, line 14: what does this file do? is it referenced anywhere else in PR?  am assuming this is one-off for yourself that doesn't need review?

soumith: it's a file that showcases how to take recorded offline trajectores, and register/visualize them.
It doesn't need to be reviewed right now.

PR #916, opened 2021-12-21
Kavya/vision error routing

kavyasrinet:
# Description

This PR introduces a new error route on top of parsing error routing. The only missing piece is retrieving the reference object from memory and double checking + adding flashing of the reference object in the renderer.

Here's what the new routing mechanism looks like:
```
"destroy the blue cube"
1. Check if the assistant did it right 
  1. Yes -> Finished take to final page 
  2. No  -> Check if this parser was right:
    1. Yes -> Check if visin model was right (extract description of reference object, check in memory and flash):
      1. Yes -> this is another error type, ask for free form description.  
      2. No -> Mark vision mistake in state
    2. No -> Mark parser mistake in state
```

Current flow:
Screen Shot 2021-12-21 at 3 06 51 PM
Screen Shot 2021-12-21 at 3 07 03 PM

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

# Testing

To repro the flow test with the command : "destroy the blue cube"
The parser gets it right but agent fails overall.

Note:
1. We still need to check the reference object against memory - this isn't trivial right now.
2. We need to add flashing of reference object to renderer.

# Checklist:

### DIFF
--- a/droidlet/dashboard/web/src/components/Interact/Question.js
+++ b/droidlet/dashboard/web/src/components/Interact/Question.js
@@ -11,15 +11,15 @@
                                /   \
                              Yes    No -> NSP Error
                               |
-                    Location Ref Exists?
+                    Reference Object Exists?
                             /   \
                           Yes    No -> Other Error
                            |
-             Found Location Ref in Memory?
+             Found Reference Object in Memory?
                          /   \
                        Yes    No -> Perception Error
                         |
-           Is 'This' the Location Ref?
+           Is 'This' the Reference Object? (point at it / flash it)
                      /   \
     Other Error <- Yes    No -> Perception Error
  */
@@ -417,22 +417,6 @@ class Question extends Component {
     this.setState({
       reference_object_description: reference_object_description,
     });
-
-    // else {
-    //   // shouldn't happen
-    //   return (
-    //     <div>
-    //       <h3> Thanks! Press to continue.</h3>
-    //       <Button
-    //         variant="contained"
-    //         color="primary"
-    //         onClick={() => this.props.goToMessage()}
-    //       >
-    //         Done
-    //       </Button>
-    //     </div>
-    //   );
-    // }
   }
 
   renderVisionQuestion() {
@@ -455,8 +439,8 @@ class Question extends Component {
         <div className="question">
           <h3>
             Okay, I was looking for an object of interest called :
-            {reference_object_description}. Here's what I think it is. Does that
-            look right ?
+            {reference_object_description}. I'll make it flash in the world now.
+            Does that object look right ?
           </h3>
           <List className="answers" component="nav">
             <ListItem button onClick={() => this.answerVision(1)}>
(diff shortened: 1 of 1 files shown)


droidlet/dashboard/web/src/components/Interact/Question.js
```diff
@@ -1,7 +1,27 @@
...
+                              |
+                    Location Ref Exists?
+                            /   \
+                          Yes    No -> Other Error
+                           |
+             Found Location Ref in Memory?
+                         /   \
+                       Yes    No -> Perception Error
+                        |
+           Is 'This' the Location Ref?
```
ethan-carlson, line 22: Maybe for clarity "Location Ref" -> "Reference Obj" since it doesn't strictly need to be in the Location subcomponent?

droidlet/dashboard/web/src/components/Interact/Question.js
```diff
@@ -337,36 +277,288 @@ class Question extends Component {
...
+    //       <Button
+    //         variant="contained"
+    //         color="primary"
+    //         onClick={() => this.props.goToMessage()}
+    //       >
+    //         Done
+    //       </Button>
+    //     </div>
+    //   );
+    // }
```
ethan-carlson, line 435: Probably don't need to preserve this, unless we think we want a different UX than people being shown a blank description in the error state.

droidlet/dashboard/web/src/components/Interact/Question.js
```diff
@@ -337,36 +277,288 @@ class Question extends Component {
...
+      // Not perception error.
+      this.setState({ view: 6 });
+      return;
+    }
+    return (
+      <div className="question">
+        <h3>
+          Okay, looks like I understood your command. I was looking for an
+          object of interest called : {reference_object_description}. Here's
+          what I think it is. Does that look right ?
```
ethan-carlson, line 453: I think we should explicitly tell people to look for the flashing object.

ethan-carlson (approved):
Overall it looks good to me; I've run it and it works for me.  Only nit is that I think we should be a bit more explicit in the vision question about what we're asking the user.

droidlet/dashboard/web/src/components/Interact/Question.js
kavyasrinet: Right yes, I did not update the comments yet, will do once I have a sign off on the implementation

droidlet/dashboard/web/src/components/Interact/Question.js
kavyasrinet: Same as above ^
This wasn't documented so I wasn't sure if I can just torch, can torch this

droidlet/dashboard/web/src/components/Interact/Question.js
kavyasrinet: Thanks, yes will do now. Flashing wasn't yet implemented

snyxan (approved):

PR #908, opened 2021-12-17
[Polymetis] Fix signal handler in RobotInterface

1heart:
# Description

- Add a test for a multi-threaded environment for RobotInterface.
- Remove signal handler

Fixes #899

## Type of change

Please check the options that are relevant.

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] Proposes a change (non-breaking change that isn't necessarily a bug)
- [ ] Refactor
- [ ] New feature (non-breaking change that adds a new functionality)
- [ ] Breaking change (fix or feature that would break some existing functionality downstream)
- [ ] This is a unit test
- [ ] Documentation only change
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

- [x] I want a thorough review of the implementation.
- [ ] I want a high level review. 
- [ ] I want a deep design review.

## Before and After

If this PR introduces a change or fixes a bug, please add before and after screenshots (if this is causes a UX change) or before and after examples(this could be plain text, videos etc ) to demonstrate the change.

If this PR introduces a new feature, please also add examples or screenshots demonstrating the feature.

# Testing

Please describe the tests that you ran to verify your changes and how we can reproduce.

# Checklist:

- [ ] I have performed manual end-to-end testing of the feature in my environment.
- [ ] I have added Docstrings and comments to the code.
- [ ] I have made changes to existing documentation where needed.
- [ ] I have added tests that show that the PR is functional.
- [ ] New and existing unit tests pass locally with my changes.
- [ ] I have added relevant collaborators to review the PR before merge.
- [ ] [Polymetis only] I have verified that all scripts in `tests/scripts` runs on hardware.

PR #926, opened 2021-12-28
[Polymetis] Fix integration tests

1heart:
## Description

I noticed the multithreading test introduced in #908 passes even before I fixed the issue. Turns out a Python process that has a thread throwing an exception still returns 0 exit code.

I also refactored & fixed a set of integration test issues which I noticed as I debugged the above.

- Fix `6_multithreaded.py` to fail appropriately when the multithreaded didn't work properly
- Split `test_integrations.sh` into multiple scripts
- None of the `.py` scripts were running because the relative path was incorrect. I changed it to be relative to the git repo root.
- A different subset of tests were running between the local and conda-pkg CI jobs. I refactored the config.yml to use a shared command called `polymetis_tests` (and had to rename the environment used by the conda-pkg tests)
- Somewhat unrelated, but removed unreferenced `create_robot_model` function from `test_utils.py`
- The franka client never exits during mocked mode when the server dies. To fix this, I throw exception when ControlUpdate fails in franka_panda_client, instead of just printing.

## Type of change

- [x] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

# Testing

I checked the fix for `6_multithreaded.py` by reverting the fix using commit `3170e32`, checking it fails appropriately in commit `daf5aea`, and reapplying the fix at commit `fe5dcd3`

# Checklist:


.circleci/config.yml
```diff
@@ -1,5 +1,43 @@
...
+          command: |
+            . /opt/conda/etc/profile.d/conda.sh
+            conda activate polymetis-local
+            ./polymetis/tests/test_mocked_hardware_client.sh
+      - run:
+          name: Run integration tests (simulation)
+          command: |
+            . /opt/conda/etc/profile.d/conda.sh
+            conda activate polymetis-local
+            ./polymetis/tests/test_simulation.sh
```
exhaustin, line 40: What's the motivation behind splitting `test_integrations.sh` into multiple scripts? I agree it looks much cleaner, but it also results in some duplicated code in the scripts and also a much longer `polymetis_tests` shared command.

exhaustin (approved):
Thanks for fixing! Looks good in general, some minor comments.

polymetis/tests/scripts/6_multithreaded.py
```diff
@@ -11,34 +11,43 @@
...
+    except Exception as e:
+        exceptions.append(e)
 
 
 if __name__ == "__main__":
     thread = threading.Thread(target=connect_and_send_policy)
     thread.start()
     thread.join()
+
+    assert success, f"Exception: {exceptions}"
```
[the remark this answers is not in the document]
1heart: Yeah, the latter doesn't work unfortunately. I was looking for a more elegant way (eg return True/False from thread) but a lot of the suggestions were "create a custom Thread object" which seemed overkill.

.circleci/config.yml
1heart: I found it exceedingly difficult to figure out what was going on when the integration tests broke -- there's a ton of printout, several error messages from killing the server repeatedly, and in general hard to pinpoint what was going on. For example, here I was testing to make sure the tests actually broke when the multithread wasn't fixed; I was expect it should only break the simulation tests.

I do agree it's repetitive -- any thoughts on how to do it while preserving clarity?

exhaustin: Some ideas:
1. Minimize printouts so as to not dilute the error messages, maybe have a consistent way of configuring verbosity so running integration tests won't have the same verbosity with a user running each process in a separate terminal.
2. Have the server exit gracefully upon an interrupt and have the test script terminate the server with signals instead of pkill-ing.
3. Pass any exception on the server side as a string field back to the user and have better error handling on the client side so it's easy to pinpoint when the exception happened from the error msgs.

Until we get there, I agree it would make things easier by keeping the scripts separate, and minor tech debt in the repetitiveness can be easily removed afterwards.

1heart: Added this as an issue in #930 for TODO post fbrp integration

PR #932, opened 2022-01-07
simple tag labels for inst seg

aszlam:
adds simple instance segmentation labels to iglu-style scene generation


aszlam:
also shifts the non-cuberite blocks over as they were shifted in cuberite

ethan-carlson (approved):
Works for me, thanks!  I have small_scenes lines 147-148 commented out on my branch right now so that the type is the same after import regardless of the number of scenes requested, but if there's a reason it's important I can add a conditional to my code.

PR #944, opened 2022-01-13
Add ignore_self to memory creation

jacklanchantin:
# Description

When searching the memory, we sometimes want to ignore the self reference object. Previously, we had an option in the forward pass, which was deprecated. This change adds the ignore_self option as an argument to the memory searcher where the default option is False.

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

# Testing

I ran the build_memory_eqa_data.py script for the neural memory repo.

# Checklist:

### DIFF
--- a/droidlet/memory/craftassist/mc_memory.py
+++ b/droidlet/memory/craftassist/mc_memory.py
@@ -63,6 +63,7 @@ class MCAgentMemory(AgentMemory):
         agent_time=None,
         coordinate_transforms=None,
         agent_low_level_data={},
+        ignore_self=False,
     ):
         super(MCAgentMemory, self).__init__(
             db_file=db_file,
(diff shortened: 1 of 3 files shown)


jacklanchantin:
@aszlam I actually don't think this is necessary. We were previously using ignore_self, but I don't think it makes a difference right now. May be useful in the future though.

droidlet/memory/craftassist/mc_memory.py
```diff
@@ -63,6 +63,7 @@ def __init__(
         agent_time=None,
         coordinate_transforms=None,
         agent_low_level_data={},
+        ignore_self=False,
```
aszlam, line 66: why do we need this here? can't we just put it in the basic_search call ?

aszlam:
If you do want to keep this, can you change it at https://github.com/facebookresearch/fairo/blob/caaa3e5541872de38e293e6495c0ce1f48869b92/droidlet/memory/memory_filters.py#L481 instead of in the agent memory?  like allow ignore_self to be passed in during a search, instead of at searcher init

jacklanchantin:
I don't think we need this as of now, so I'm going to close this PR.

PR #957, opened 2022-01-25
[Polymetis] Rename RobotInterface API

exhaustin:
# Description

A more systematic naming scheme for `RobotInterface` methods.

Retains backward compatibility by having old methods reference new methods.

## Type of change

- [x] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

# Testing

Does not replace function names in tests, ensuring backward compatibility.

# Checklist:

- [ ] [Polymetis only] I ran on hardware (1) that all scripts in `tests/scripts`, (2) asv benchmarks.


polymetis/polymetis/python/polymetis/robot_interface.py
```diff
@@ -459,17 +472,34 @@ def set_ee_pose(
 
         return self.send_torch_policy(torch_policy=torch_policy, **kwargs)
 
+    """
+    PyRobot backward compatibility methods
+    """
```
1heart, line 477: - Print an error message saying these are deprecated? e.g. `log.warn` and suggest the new method name
- Add docstring with bolded deprecation warning (and then make sure to regenerate the documentation)

polymetis/polymetis/python/polymetis/robot_interface.py
```diff
@@ -430,15 +429,29 @@ def set_ee_pose(
...
+            ee_quat_desired = ee_quat_current
+        else:
+            assert (
+                len(orientation) == 4
+            ), "Only quaternions are accepted as orientation inputs."
+            ee_quat_desired = torch.Tensor(orientation)
+            if delta:
+                ee_quat_desired = (
+                    R.from_quat(ee_quat_current) * R.from_quat(ee_quat_desired)
+                ).as_quat()
```
1heart, line 447: Is this the right order of applying rotations? Shouldn't ee_quat_desired be applied onto the current position?

More generally, can you add unit tests with `delta=True` so we're sure this new functionality is working?

1heart (changes requested):
Thanks for doing this -- some cleanup/documentation/testing would be good.

Also, I don't believe the scripts under `polymetis/tests/scripts` are updated with the new method names. Would you mind find/replace throughout the repo (including the documentation) to make sure these changes are updated?

exhaustin:
> Thanks for doing this -- some cleanup/documentation/testing would be good.
> 
> Also, I don't believe the scripts under `polymetis/tests/scripts` are updated with the new method names. Would you mind find/replace throughout the repo (including the documentation) to make sure these changes are updated?

The test scripts are intentionally unchanged -- so we can verify that everything that works before can still work.
However, now that tests are working I guess I can change them now even though that would be two separate incremental changes in a PR.

polymetis/polymetis/python/polymetis/robot_interface.py
exhaustin: Should we be deprecating them right away if we're still advocating Polymetis as a drop-in replacement for PyRobot?
It doesn't make sense if a user replaces PyRobot calls with Polymetis and ends up with a bunch of deprecation messages.
However, if not many people are affected then I think we can deprecate them. Do you know if there are any modules (maybe in droidlet) that switched to Polymetis from PyRobot?

1heart: Droidlet has not switched from PyRobot to Polymetis, though tbh PyRobot is unmaintained and doesn't seem to have been touched since Zisu left. You can ping the droidlet team to be sure, but in general I am fine with adding deprecation messages.

exhaustin: Sounds good, I'll just deprecate them.

exhaustin: Also, I thought the documentation auto-regenerate & upload in CircleCI?

1heart (approved):
Thanks!

polymetis/polymetis/python/polymetis/robot_interface.py
1heart: Not the manually-written rst files -- not sure if any needs to be updated.

PR #959, opened 2022-01-27
[Polymetis] Add impedance control macros to RobotInterface

exhaustin:
# Description

Add macros for updatable impedance control (joint & cartesian) to RobotInterface.

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

# Testing

Added tests in `test/scripts`

# Checklist:

- [x] [Polymetis only] I ran on hardware (1) that all scripts in `tests/scripts`, (2) asv benchmarks.


exhaustin:
Concerns:
- Do we need safety measures that prevents users from updating the desired state with something very far away from current states?
- Should we check whether the server is still running the impedance controller when updating? Pros: Handles cases where the impedance controller is terminated due to communication constraints, safety limits, other server issues, or another client overriding a command; Cons: Much updates due to an additional RPC call.

polymetis/polymetis/python/polymetis/robot_interface.py
```diff
@@ -456,22 +461,104 @@ def move_to_ee_pose(
...
+
+        param_dict = {}
+        if position is not None:
+            param_dict["ee_pos_desired"] = position
+        if orientation is not None:
+            param_dict["ee_quat_desired"] = orientation
+
+        return self.update_current_policy(param_dict)
+
+    def terminate_impedance_controller(self):
```
1heart, line 554: Do we need a separate of impedance controllers specifically?

1heart (commented):
IMO (1) is not within the scope of this PR, though it may be nice to have in the future.

(2) should be covered on the server side.

In general, I would prefer to have the check for whether a command be validated not through an `enum` on the client side (which could easily fall out of sync with the reality on the server), but rather from the server providing the source of truth (e.g. which type of controller is being run) and validating the client's attempted actions.

That doesn't mean that we need a separate RPC call to validate each `update` command -- I actually think (correct me if I'm wrong) an exception is already thrown if you attempt to update a field which doesn't exist. That exception can be surfaced on the client side when you attempt to update a controller incorrectly.

polymetis/polymetis/python/polymetis/robot_interface.py
exhaustin: ?

1heart: Separate termination of impedance controllers*

exhaustin: I don't think there needs to be. I think even having a single termination method for all non-blocking controllers would've been fine if we don't need to keep track of the mode on client side.

exhaustin:
Hmm here's what happens currently if you try to update the server with a field that doesn't exist: The client side catches the exception but still freezes. The server doesn't print out any error messages but is unable to accept any other command, even after restarting the client.
I will try to fix the controller update error handling in a separate PR.

1heart (approved):
Great, thank you!

PR #967, opened 2022-02-02
[Polymetis] Fix controller update error handling

exhaustin:
# Description

- Fix issue where mutex doesn't unlock when an error is raised during controller update, causing the server to hang in a state where it's unable to accept any other connections.
- Forward error messages from running torchscript modules to client.
- Change `RobotInterface` to directly raise gRPC errors.
- Initialize returned interval indexes so it returns -1 instead of a random uninitialized number under the event of an error.

## Type of change

- [ ] Refactor
- [x] Breaking change (fix or feature that would break some existing functionality downstream)
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

Previously updating the controller with an erroneous field would result in the client side receiving an uninformative error:
image
while nothing gets printed out on the server side. Any client is then unable of connecting to the server due to invalid controller context:
image

After the fix, the client throws the appropriate exception:
image
while the server ignores the erroneous update and continues to function normally.

# Testing

# Checklist:

- [ ] [Polymetis only] I ran on hardware (1) that all scripts in `tests/scripts`, (2) asv benchmarks.

### DIFF
--- a/polymetis/polymetis/src/polymetis_server.cpp
+++ b/polymetis/polymetis/src/polymetis_server.cpp
@@ -292,17 +292,19 @@ Status PolymetisControllerServerImpl::UpdateController(
 
   // Update controller & set intervals
   if (custom_controller_context_.status == RUNNING) {
-    bool success;
-    std::string error_msg;
-
     custom_controller_context_.controller_mtx.lock();
     interval->set_start(robot_state_buffer_.size());
-    success =
-        custom_controller_context_.custom_controller->param_dict_update_module(
-            error_msg);
+    int code = custom_controller_context_.custom_controller
+                   ->param_dict_update_module();
     custom_controller_context_.controller_mtx.unlock();
 
-    if (!success) {
+    if (code) {
+      std::string error_msg;
+      if (code == 1) {
+        error_msg = "error updating controller: Invalid parameter name.";
+      } else if (code == 2) {
+        error_msg = "error updating controller: Tensor shape mismatch.";
+      }
       return Status(StatusCode::CANCELLED, error_msg);
     }
 
--- a/polymetis/polymetis/src/polymetis_server.cpp
+++ b/polymetis/polymetis/src/polymetis_server.cpp
@@ -292,10 +292,20 @@ Status PolymetisControllerServerImpl::UpdateController(
 
   // Update controller & set intervals
   if (custom_controller_context_.status == RUNNING) {
+    bool success;
+    std::string error_msg;
+
     custom_controller_context_.controller_mtx.lock();
     interval->set_start(robot_state_buffer_.size());
-    custom_controller_context_.custom_controller->param_dict_update_module();
+    success =
+        custom_controller_context_.custom_controller->param_dict_update_module(
+            error_msg);
     custom_controller_context_.controller_mtx.unlock();
+
+    if (!success) {
+      return Status(StatusCode::CANCELLED, error_msg);
+    }
+
   } else {
     spdlog::warn(
         "Tried to perform a controller update with no controller running.");
--- a/polymetis/polymetis/src/polymetis_server.cpp
+++ b/polymetis/polymetis/src/polymetis_server.cpp
@@ -222,6 +222,9 @@ Status PolymetisControllerServerImpl::SetController(
     LogInterval *interval) {
   std::lock_guard<std::mutex> service_lock(service_mtx_);
 
+  interval->set_start(-1);
+  interval->set_end(-1);
+
   // Read chunks of the binary serialized controller. The binary messages
   // would be written into the preallocated buffer used for the Torch
   // controllers.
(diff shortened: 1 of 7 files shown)


polymetis/polymetis/src/polymetis_server.cpp
```diff
@@ -294,8 +294,20 @@ Status PolymetisControllerServerImpl::UpdateController(
...
+
+    if (code) {
+      std::string error_msg;
+      if (code == 1) {
+        error_msg = "error updating controller: Invalid parameter name.";
+      } else if (code == 2) {
+        error_msg = "error updating controller: Tensor shape mismatch.";
+      }
+      return Status(StatusCode::CANCELLED, error_msg);
+    }
```
1heart, line 309: My thought is that it's hard to exhaustively enumerate all possible exceptions that could possibly happen while calling param_dict_update_module. Instead, can we wrap this in a try/catch, same as e.g. line 237-257, and directly return the error message?

exhaustin: These are the most common error messages that users might run into. If we're giving users the freedom to update the controller with any dictionary, I think it would be very helpful to include more descriptive error messages.

1heart: Yes, but aren't the error messages from catching the exception the same as the ones you're manually writing here?

stuart-fb:
"Fixes issue where mutex doesn't unlock when an error is raised during controller update, causing the server" - I think there's more to this sentence?  Assuming the server hangs.

polymetis/polymetis/torch_isolation/src/torch_server_ops.cpp
```diff
@@ -171,8 +171,8 @@ bool TorchScriptedController::param_dict_load(char *data, size_t size) {
   return true;
 }
 
-void TorchScriptedController::param_dict_update_module() {
-  module_->data.get_method("update")(param_dict_input_->data);
+int TorchScriptedController::param_dict_update_module() {
```
stuart-fb, line 174: If extensibility is a question here, maybe return either an enum or a struct that contains an optional error string and a status enum?

polymetis/polymetis/src/polymetis_server.cpp
exhaustin: I'll investigate how informative the error messages are if we allow torchscript modules to error out, or if we're able to raise custom exceptions from within torchscript modules.

Right now the torchscript controller module is designed to check for these user input induced errors, ignore the input if something isn't right and return an error code to ensure that the server doesn't crash. So given that torchscript module design enumeration of the error code has to happen somewhere, and I chose to enumerate it outside of the torchscript isolated module so as to get access to c++ stuff such as std strings.

1heart: I believe the torchscript error just gets propagated as an `std::exception` and you can forward `e.what()` to the user client -- please lmk if that's not the case

polymetis/polymetis/torch_isolation/src/torch_server_ops.cpp
exhaustin: Yixin's suggestion of forwarding the error message from within the torchscript controller worked pretty nicely (see updated "Before and After" screenshot).
Now I'm thinking of condensing the duplicated debug_error_string in the traceback so it's easier to read.

polymetis/polymetis/src/polymetis_server.cpp
```diff
@@ -292,17 +295,28 @@ Status PolymetisControllerServerImpl::UpdateController(
...
+
   } else {
-    spdlog::warn(
-        "Tried to perform a controller update with no controller running.");
-    interval->set_start(-1);
-  }
+    std::string error_msg =
+        "Tried to perform a controller update with no controller running.";
+    spdlog::warn(error_msg);
```
1heart, line 316: Relatedly & in response to Giri's request, I think this and line 340 should actually throw on the clientside & interrupt

exhaustin:
One concern after the changes: The way we're handling errors within `RobotInterface` seems a little messy. Should we wrap all gRPC requests in a function call that handles errors in the same way?

1heart (approved):
Re: messiness, we could append "Polymetis server error\n" to the messages on the server, and then just not wrap the gRPC calls on the client. But I'm not too worried about this.

Thanks for the fix!

exhaustin:
> Re: messiness, we could append "Polymetis server error\n" to the messages on the server, and then just not wrap the gRPC calls on the client. But I'm not too worried about this.

We still need to catch the exceptions on the RobotInterface side for cleaner debug messages (without the `debug_error_string`), so I'll just leave it as it for now. I agree it's fairly minor and we can worry about it later if it proves to be an issue.

PR #978, opened 2022-02-04
mobile app initial revision

ys1382fb:
# Description

Mobile app that allows the user to label an item in AR and save the result to the cloud.

The significant files are the Swift source code. Even if you are new to Swift, it should be readable. The most important ones are:
- ARViewContainer.swift, which has the augmented reality UI code. It has some comments describing specifically what it does with AR. Mostly it calls iOS AR APIs.
- Firebase.swift, which saves user-annotation data to the cloud.

There are a number of .png files, which are just app icons of different sizes.

There is the Resnet50 image classification model, which is 100MB and not included.

There is a Google plist file for Firebase access, which is also not included, for security reasons.

## Type of change

- [ ] Refactor
- [X] New feature (non-breaking change that adds a new functionality)
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

- [X] I want a high level review. 

# Checklist:

https://user-images.githubusercontent.com/87049088/152593440-770cf8bb-0dd0-43f6-b637-fd639fbee3fc.mp4


soumith:
It's worth checking in the binary assets into the Amazon s3 bucket.
We're not going to do well by checking in all our binary assets into the git repo.

soumith (changes requested):
as commented above

ys1382fb:
> as commented above

Thanks for the review. I removed the ML model, as well as the Google Firebase access plist. The app icons I kept in the commit, as this is standard practice for a mobile app.

PR #1066, opened 2022-03-29
Add depth to realsense_wrapper

1heart:
# Description

Add simple depth images and pointclouds to realsense wrapper. Makes sure to align the depth images.

Note -- is there a simple testing harness for the realsense wrapper?

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

# Testing

# Checklist:

### DIFF
--- a/perception/realsense_driver/python/realsense_wrapper/realsense_d435.py
+++ b/perception/realsense_driver/python/realsense_wrapper/realsense_d435.py
@@ -53,19 +53,23 @@ class RealsenseAPI:
 
     def get_images(self, depth=False):
         framesets = self._get_frames()
+
+        align_to = rs.stream.color
+        align = rs.align(align_to)
+        aligned_framesets = [align.process(frameset) for frameset in framesets]
+
         imgs = []
         depth_imgs = []
-        for frameset in framesets:
+        for frameset in aligned_framesets:
             frame = frameset.get_color_frame()
             img = np.asanyarray(frame.get_data())
+            imgs.append(img)
 
             if depth:
                 frame = frameset.get_depth_frame()
                 depth_img = np.asanyarray(frame.get_data())
                 depth_imgs.append(depth_img)
-
-            imgs.append(img)
-
+            
         if depth:
             return imgs, depth_imgs
         else:
--- a/perception/realsense_driver/python/realsense_wrapper/realsense_d435.py
+++ b/perception/realsense_driver/python/realsense_wrapper/realsense_d435.py
@@ -5,7 +5,7 @@ import pyrealsense2 as rs
 
 
 class RealsenseAPI:
-    """ Wrapper that implements boilerplate code for RealSense cameras """
+    """Wrapper that implements boilerplate code for RealSense cameras"""
 
     def __init__(self):
         # Identify devices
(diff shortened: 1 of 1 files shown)


perception/realsense_driver/python/realsense_wrapper/realsense_d435.py
```diff
@@ -51,12 +51,38 @@ def get_intrinsics(self):
     def get_num_cameras(self):
         return len(self.device_ls)
 
-    def get_images(self):
+    def get_images(self, depth=False):
```
0wu, line 54: should we consider having `get_depth()` similar to `get_point_clould()` below?

perception/realsense_driver/python/realsense_wrapper/realsense_d435.py
```diff
@@ -51,12 +51,38 @@ def get_intrinsics(self):
...
+        else:
+            return imgs
+
+    def get_pointcloud(self) -> rs.pointcloud:
+        framesets = self._get_frames()
+        pc_ls = []
+        for frameset in framesets:
+            pc = rs.pointcloud()
+            pc.map_to(frameset.get_color_frame())
+            points = pc.calculate(frameset.get_depth_frame())
```
0wu, line 85: should we use aligned depth frame here?

1heart:
Will wait for @0wu's refactor

perception/realsense_driver/python/realsense_wrapper/realsense_d435.py
exhaustin, line 54: 1. The process for retrieving rgb and depth images are quite independent from each other. Might be easier to just have two separate APIs (`get_rgb_images` and `get_depth_images`), especially since having `depth=True` changes the number of output arguments in the current implementation.
2. A lot of duplicated code between retrieving rgb images and depth images. Maybe we can have a base method `_get_images(self, type=RGB / DEPTH)` (with enums for RGB and DEPTH)?

perception/realsense_driver/python/realsense_wrapper/realsense_d435.py
exhaustin: In some cases we want the raw point cloud without interpolations, and in other cases we want to align the points within the point cloud to pixels in the rgb image. Since this is meant to be a general API, we should probably support both.

exhaustin (changes requested):
Side note: I'm starting to think that returning a list of outputs for each camera might be confusing at times, and also I'm worried that there isn't a guaranteed order for the cameras to be detected between runs. Thoughts on having outputs take the form of a dictionary with device IDs as keys?

perception/realsense_driver/python/realsense_wrapper/realsense_d435.py
1heart: I think the main thing is that we want the RGB + D images to correspond to each other, which calling multiple methods may not result in. Discussing with @0wu we decided on just returning a whole named tuple of [rgb, d, pointcloud] which is aligned & sync'd.

perception/realsense_driver/python/realsense_wrapper/realsense_d435.py
1heart: Yes, probably should support both

perception/realsense_driver/python/realsense_wrapper/realsense_d435.py
```diff
@@ -51,12 +56,25 @@ def get_intrinsics(self):
...
+        """Returns a numpy array of [n_cams, height, width, RGBD]"""
         framesets = self._get_frames()
-        imgs = []
-        for frameset in framesets:
-            frame = frameset.get_color_frame()
-            img = np.asanyarray(frame.get_data())
-            imgs.append(img)
+        num_cams = self.get_num_cameras()
+
+        rgbd = np.zeros([num_cams, self.height, self.width, 4], dtype=np.uint16)
```
0wu, line 64: minor - consider using np.ndarray to save the time zeroing the array as the values are going to be overwritten anyways.

0wu (approved):
thanks for the change.
LGTM.

exhaustin (approved):
Looks great!

Minor concern: Are we going to have intended usecases where we're using cameras of different resolutions and/or fps? Slightly worried since cameras are a scarce resource right now so we might not get enough quantities of the exact same camera, but if it's not a problem for the time being this interface is pretty clean :)

1heart:
I totally agree the realsense_wrapper isn't the most general interface. I'm not sure which cameras we have in the pipeline, but with the barebones/simple nature of the wrapper I'm hoping it becomes easy to extend when we get there.

perception/realsense_driver/python/realsense_wrapper/realsense_d435.py
1heart: `np.empty` right? will do in a future PR

PR #1077, opened 2022-04-05
New Agent Chat Schema

ethan-carlson:
# Description

This is a PR to discuss the adoption of a JSON schema that will govern chats sent from the agent to the user.  It should support, or provide a platform to build on for ongoing workstreams in human-in-the-loop, multi-turn dialog, mobile AR, and multi-agent swarms.

The file being checked in is a single piece of documentation, so please see that file for more details.

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

- [X] I want a deep design review.

## Before and After

Previously chats sent from the agent take the form of plain text strings without any metadata or hints as to their purpose / what should be done with them.  This is fine for the existing use case where the agent most often issues chats that don't require response from the user or special rendering.  The new schema should provide a structure for those new use cases.

# Checklist:

- [X] I have made changes to existing documentation where needed.

### DIFF
--- a/droidlet/dialog/chat_schema.md
+++ b/droidlet/dialog/chat_schema.md
@@ -1,0 +1,71 @@
+## Description ##
+This is the source of truth for the most updated version of the schema to be followed when sending chats from the agent the the user (and perhaps eventually from the agent to other agents).  When updating this document please increment the version number.  Older versions of the schema can be found in git history.
+
+Note that chats sent from the user to the agent should always be in plaintext, as all messages should be parsed by the NSP and interpreted.  I.e. special pathways and conditionals on agent input are discouraged.  It is possible to help the NSP by including special tokens that are reserved for specific use cases, eg. clarification response.
+
+### Content Types ###
+ - `chat_string`: Only text to display in the chat window with no response assumed.
+    - Example: "I finished building this!"
+ - `chat_and_media`: Text and media to display in the chat window, with no response assumed.
+    - Example: "I finished building this!" (with accompanying photo)
+ - `chat_and_text_options`: Text to display in the chat window along with a list of possible response options.
+    - Example: "Are these the Droidlets you are looking for?" ('no', 'no')
+ - `chat_and_media_options`: Text to display in the chat window along with a list of clickable media options.
+    - Example: "Which of these is the 'blue sphere' you were referring to?" (pictures of blue spheres)
+ - `chat_and_media_and_text_options`: Text and media to display in the chat window alond with a list of possible response options.
+    - Example: "Is this funny?" <img src='https://external-preview.redd.it/hWh_8TpqrT6zAwpzHJ_m9Rx3iHjc_yI4zSI6aazMFTc.jpg?auto=webp&s=6d8006ac3edca5bad98dd7b5b9a4a8d5554eaff0' width="200"> ('yes', 'no')
+
+### Schema ###
+
+<pre>
+{
+	"title": "Droidlet Chat",
+	"description": "A single chat sent from a Droidlet agent to the user",
+    "version": 1,
+	"type": "object",
+	
+	"properties": {
+		"speaker_id": {
+			"description": "The memid of the speaker",
+			"type": "string"
+		},
+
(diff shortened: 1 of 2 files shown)


droidlet/dialog/chat_schema.md
```diff
@@ -0,0 +1,71 @@
...
+ - `chat_and_media_options`: Text to display in the chat window along with a list of clickable media options.
+    - Example: "Which of these is the 'blue sphere' you were referring to?" (pictures of blue spheres)
+ - `chat_and_media_and_text_options`: Text and media to display in the chat window alond with a list of possible response options.
+    - Example: "Is this funny?" <img src='https://external-preview.redd.it/hWh_8TpqrT6zAwpzHJ_m9Rx3iHjc_yI4zSI6aazMFTc.jpg?auto=webp&s=6d8006ac3edca5bad98dd7b5b9a4a8d5554eaff0' width="200"> ('yes', 'no')
+
+### Schema ###
+
+<pre>
+{
+	"title": "Droidlet Chat",
```
kavyasrinet, line 22: to avoid confusion : `"title": "Droidlet agent initiated chat"`

droidlet/dialog/chat_schema.md
```diff
@@ -0,0 +1,71 @@
...
+	"description": "A single chat sent from a Droidlet agent to the user",
+	"version": 1,
+	"type": "object",
+
+	"properties": {
+		"speaker_id": {
+			"description": "The memid of the speaker",
```
kavyasrinet, line 29: isn't the speaker going to be the agent here ?

ethan-carlson: Yes, this field may be superfluous now.  I included it for two reasons.  One is that speaker memid, chat text, and timestamp are the three ChatNode arguments, so having the memid accessible could make for convenient retrieval of the memory node.  Though we could presumably get it from the agent at pretty much any time anyway.  The second reason is looking forward to a situation in which more speakers are around than just the agent and the user.  But it's understandable if you want to plan for this use case when it comes around.

Let me know if this is enough justification to keep it or if you'd prefer to drop it for now.

droidlet/dialog/chat_schema.md
```diff
@@ -0,0 +1,71 @@
...
+			"type": "string"
+		},
+
+		"timestamp": {
+			"description": "Unix timestamp of when the chat was sent",
```
ys1382fb, line 34: I assume you mean UTC and milliseconds, but best to say explicitly.

droidlet/dialog/chat_schema.md
```diff
@@ -0,0 +1,71 @@
...
+					"chat_and_text_options",
+					"chat_and_media_options",
+					"chat_and_media_and_text_options"
+				]
+			},
+			"minItems": 1,
+			"maxItems": 1
+		},
+
+		"content": {
```
ys1382fb, line 54: We should not limit the response options to 2. There are cases where the user may have to select from a larger number of options.

droidlet/dialog/chat_schema.md
ethan-carlson: I replaced this with the memid of the chat itself.

droidlet/dialog/chat_schema.md
ethan-carlson: Since the inner array is specified exactly the min/max 2 is actually irrelevant, I think, so I deleted it for clarity.  However just so we're in the same page, the format of this field is an array of arrays where the outer array has length >= 1 and the inner arrays consist of two items, essentially a key:value pair.  Here's an example of what the content field could look like:

`[ ["text", "Is this the cube you want me to destroy?"], ["response_option", "yes"], ["response_option", "no"] ]`

ys1382fb: Thanks for the clarification. In that case, why not use a JSON Object (ie KVP) instead of an array?

kavyasrinet (approved):

PR #1083, opened 2022-04-06
set get_all true. return multiple values

jacklanchantin:
# Description

This allows us to return all of the property values for a `memory.search()`

Previously in `handle_output()`, `get_property_value()` had `get_all=False`, so it was only returning one property ("mob") for every tag query. This PR sets `get_all=True` in `get_propert_value()`.

If `get_all=False`, `get_propert_value()` returns a string, so I first check if the return is a list. Each returned value in the list is added to the dict `values_dict` for the corresponding memid key. Finally, all values for all memid keys are concatenated into a list and returned in `handle_output()`.

Note: this now returns a list of lists (one list for each memid).

Fixes # (issue)

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

Before:
```
print(self.handle_output(agent_memory, query, memids))
['mob']
```

After:
```
print(self.handle_output(agent_memory, query, memids))
[['mob', '_physical_object', '_animate', '_not_location', 'rabbit', 'gray']]
```

# Testing

# Checklist:


jacklanchantin:
@aszlam `get_property_value()` is also called in `handle_selector()`. Should we set `get_all=True` there too?

droidlet/memory/memory_filters.py
```diff
@@ -471,10 +471,12 @@ def handle_output(self, agent_memory, query, memids):
                         )
                     )
                 for m in memids:
-                    values_dict[m].append(get_property_value(agent_memory, m, aname))
-            if len(attribute_name_list) == 1:
-                for m in values_dict:
-                    values_dict[m] = values_dict[m][0]
+                    prop_vals = get_property_value(agent_memory, m, aname, get_all=True)
```
aszlam, line 474: this should be propagated from .search().    To be backwards compatible, make the default False for now; and in the nsm code use .search(query, get_all=True)

droidlet/memory/memory_filters.py
```diff
@@ -471,10 +471,12 @@ def handle_output(self, agent_memory, query, memids):
...
+                    if type(prop_vals) is list:
+                        for prop_val in prop_vals:
+                            values_dict[m].append(prop_val)
+                    else:
+                        values_dict[m].append(prop_vals)
             # TODO switch everything to dicts
             return [values_dict[m] for m in memids]
```
aszlam, line 481: this needs to be finessed- we should probably be returning a separate record for each value.  so if a memid m has multiple associated properties p0, p1, p2, we should eventually return (from .search()) [m, m, m], [p0, p1, p2]

jacklanchantin: @aszlam see recent commits. This format breaks crafassist unit test when the vals are (x,y) coords:

 ```
query = "SELECT (x, y) FROM ReferenceObject WHERE ((has_tag=plays_volleyball) AND (x<0))"
memids, vals = m.search(self.memory, query=query)
>       assert abs(vals[0][0] + 2.0) < 0.01
E       TypeError: 'float' object is not subscriptable
```

I need to think about how to fix this.

File: droidlet/memory/memory_filters.py [hunk 3]
```diff
@@ -463,22 +463,31 @@ def handle_output(self, agent_memory, query, memids):
...
+                        for prop_val in prop_vals:
+                            attributes.append(prop_val)
+                    else:
+                        attributes.append(prop_vals)
+                if len(attribute_name_list) == 1:
+                    # attributes for this memid are single values
+                    values_dict[m] += attributes
+                else:
+                    # attributes for this memid are a list
+                    values_dict[m].append(attributes)
```
jacklanchantin, line 486: I think this is correct. If there are multiple attribute names, then the props are a list of lists (one for each attribute name). For example, (x,y) coords have two attributes in attribute_name_list: [x,y].

aszlam (approved):

PR #1142, opened 2022-05-11
[Hello] Add optional ROS driver, hook up Hector SLAM for base odometry

soumith:
## Effects: 

This PR brings in
-  significantly improved base odometry powered by a global LIDAR scan matcher, combined with a global optimizer that accounts for loop closures
- fixes some bugs in the planner that were making it stuck in-place in rare situations, because the grid map resolution, combined with navigation thresholds made it loop indefinitely

## Infrastructure

This PR adds:

- A ROS driver to
  - the base and joint {movement and statuses}
  - the lidar scans
  - [Hector SLAM](http://wiki.ros.org/hector_slam), a lidar-based SLAM with loop closure

The ROS driver was added for the base/joint/lidar because ROS generally wants to take full control of sensors -- and Hector SLAM is only available via ROS. The base/joint/lidar APIs dont really allow simultaneous usage among two processes.

The ROS driver is fully consistent with the non-ROS API. Userland should notice no difference.

### First commit: ROS drivers

You would want to review the first commit first, it's a large meaty commit.

- `launch.sh` adds in launching of ROS or non-ROS codepath via the `./launch.sh --ros` command-line argument
  - If running the ROS codepath, the user is expected to run `./launch_ros.sh` before-hand in a different terminal (as documented in the README). This launches some of the [stretch_core](https://github.com/hello-robot/stretch_ros/tree/master/stretch_core) ROS Services, and then launches hector_slam ROS Services.
- The ROS driver changes are isolated into four files (apart from `.launch` files):
  - `remote_hello_robot_ros.py` - this gives an API-compatible services as `remote_hello_robot.py`, but in the backend uses the ROS driver
  - `lidar_ros_driver.py` - this is the ROS driver that exposes lidar scans in the exact same format as original driver `lidar.py`
  - `stretch_ros_move_api.py` - this is the ROS driver that exposes move, base odometry, joint states, slam odometry
- `remote_hello_realsense.py` now has an optional `--ros` argument, which enables the ROS codepath for pulling LIDAR scans (and hence imports and uses `lidar_ros_driver.py` if this option is enabled)
- `goto` in `utils.py` which executes base movement for straightline trajectories has been generalized to work for both ROS and non-ROS backends

Other notes for this commit:
- We use the defaults for Hector SLAM (stretch_hector*.launch files), not trying to tune things yet. The defaults work great.

### Second commit:

- A subtle type-promotion bug identified during operation. If you give the XYT to `transform_global_to_base` as `[0, 0, 0]` then it treats it as a numpy int array, and the output transform ends up being wrong. The fix is simple, always cast it to float32 on entry

### Third commit:

The planner has lower resolution than the real world. By default, it has a planning grid with 5cm x 5cm tiles.
So, When the planner gets a given goal in world co-ordinates, it pulls a short-term goal, and it checks if the goal is within planner's resolution bounds.
However, this bounds-checking was incorrect because it used L2 Distance over the side of the grid, didn't account for diagonal / hypotenuse cases. I just simplified it by checking individual dimensions separately.

### DIFF
--- a/droidlet/lowlevel/hello_robot/hello_robot_mover.py
+++ b/droidlet/lowlevel/hello_robot/hello_robot_mover.py
@@ -217,7 +217,7 @@ class HelloRobotMover(MoverInterface):
         """reset the camera to 0 pan and tilt."""
         return self.bot.reset()
 
-    def move_relative(self, xyt_positions, blocking=True):
+    def move_relative(self, xyt_positions, blocking=True, distance_threshold=None, angle_threshold=None):
         """Command to execute a relative move.
 
         Args:
@@ -229,11 +229,11 @@ class HelloRobotMover(MoverInterface):
             xyt_positions = [xyt_positions]
         for xyt in xyt_positions:
             self.nav_result.wait()
-            self.nav_result = safe_call(self.nav.go_to_relative, xyt)
+            self.nav_result = safe_call(self.nav.go_to_relative, xyt, 10000000000, distance_threshold, angle_threshold)
             if blocking:
                 self.nav_result.wait()
 
-    def move_absolute(self, xzt_positions, blocking=True):
+    def move_absolute(self, xzt_positions, blocking=True, distance_threshold=None, angle_threshold=None):
         """Command to execute a move to an absolute position.
 
         It receives positions in canonical world coordinates and converts them to pyrobot's coordinates
@@ -250,7 +250,7 @@ class HelloRobotMover(MoverInterface):
             logging.info("Move absolute in canonical coordinates {}".format(xzt))
             self.nav_result.wait()
             robot_coords = base_canonical_coords_to_pyrobot_coords(xzt)
-            self.nav_result = self.nav.go_to_absolute(robot_coords)
+            self.nav_result = self.nav.go_to_absolute(robot_coords, 10000000000, distance_threshold, angle_threshold)
             if blocking:
                 self.nav_result.wait()
         return "finished"
--- a/droidlet/lowlevel/hello_robot/hello_robot_mover.py
+++ b/droidlet/lowlevel/hello_robot/hello_robot_mover.py
@@ -225,6 +225,14 @@ class HelloRobotMover(MoverInterface):
         Args:
             xzt_positions: a list of relative (x,y,yaw) positions for the bot to execute.
             x,y,yaw are in the pyrobot's coordinates.
+            distance_threshold: the distance resolution (in metres) for the move to be deemed successful
+                                If distance_threshold=0.1, and you ask the robot to move 0.5, then
+                                if the robot moves 0.4 or 0.6, it still
+                                exits the navigation with a success
+            angle_threshold: the angle resolution (in degrees) for the move to be deemed successful
+                                If angle_threshold=5, and you ask the robot to rotate 30 degrees, then
+                                if the robot rotates 25 or 35 degrees, it still
+                                exits the navigation with a success
         """
         if not isinstance(next(iter(xyt_positions)), Iterable):
             # single xyt position given
@@ -232,7 +240,9 @@ class HelloRobotMover(MoverInterface):
         for xyt in xyt_positions:
             self.nav_result.wait()
             self.nav_result = safe_call(
-                self.nav.go_to_relative, xyt, 10000000000, distance_threshold, angle_threshold
+                self.nav.go_to_relative, goal=xyt,
+                distance_threshold=distance_threshold,
+                angle_threshold=angle_threshold
             )
             if blocking:
                 self.nav_result.wait()
@@ -257,7 +267,9 @@ class HelloRobotMover(MoverInterface):
             self.nav_result.wait()
             robot_coords = base_canonical_coords_to_pyrobot_coords(xzt)
             self.nav_result = self.nav.go_to_absolute(
-                robot_coords, 10000000000, distance_threshold, angle_threshold
+                goal=robot_coords,
+                distance_threshold=distance_threshold,
+                angle_threshold=angle_threshold
             )
             if blocking:
                 self.nav_result.wait()
--- a/droidlet/lowlevel/hello_robot/hello_robot_mover.py
+++ b/droidlet/lowlevel/hello_robot/hello_robot_mover.py
@@ -240,9 +240,10 @@ class HelloRobotMover(MoverInterface):
         for xyt in xyt_positions:
             self.nav_result.wait()
             self.nav_result = safe_call(
-                self.nav.go_to_relative, goal=xyt,
+                self.nav.go_to_relative,
+                goal=xyt,
                 distance_threshold=distance_threshold,
-                angle_threshold=angle_threshold
+                angle_threshold=angle_threshold,
             )
             if blocking:
                 self.nav_result.wait()
@@ -269,7 +270,7 @@ class HelloRobotMover(MoverInterface):
             self.nav_result = self.nav.go_to_absolute(
                 goal=robot_coords,
                 distance_threshold=distance_threshold,
-                angle_threshold=angle_threshold
+                angle_threshold=angle_threshold,
             )
             if blocking:
                 self.nav_result.wait()
--- a/droidlet/lowlevel/hello_robot/hello_robot_mover.py
+++ b/droidlet/lowlevel/hello_robot/hello_robot_mover.py
@@ -217,7 +217,9 @@ class HelloRobotMover(MoverInterface):
         """reset the camera to 0 pan and tilt."""
         return self.bot.reset()
 
-    def move_relative(self, xyt_positions, blocking=True, distance_threshold=None, angle_threshold=None):
+    def move_relative(
+        self, xyt_positions, blocking=True, distance_threshold=None, angle_threshold=None
+    ):
         """Command to execute a relative move.
 
         Args:
@@ -229,11 +231,15 @@ class HelloRobotMover(MoverInterface):
             xyt_positions = [xyt_positions]
         for xyt in xyt_positions:
             self.nav_result.wait()
-            self.nav_result = safe_call(self.nav.go_to_relative, xyt, 10000000000, distance_threshold, angle_threshold)
+            self.nav_result = safe_call(
+                self.nav.go_to_relative, xyt, 10000000000, distance_threshold, angle_threshold
+            )
             if blocking:
                 self.nav_result.wait()
 
-    def move_absolute(self, xzt_positions, blocking=True, distance_threshold=None, angle_threshold=None):
+    def move_absolute(
+        self, xzt_positions, blocking=True, distance_threshold=None, angle_threshold=None
+    ):
         """Command to execute a move to an absolute position.
 
         It receives positions in canonical world coordinates and converts them to pyrobot's coordinates
@@ -250,7 +256,9 @@ class HelloRobotMover(MoverInterface):
             logging.info("Move absolute in canonical coordinates {}".format(xzt))
             self.nav_result.wait()
             robot_coords = base_canonical_coords_to_pyrobot_coords(xzt)
-            self.nav_result = self.nav.go_to_absolute(robot_coords, 10000000000, distance_threshold, angle_threshold)
+            self.nav_result = self.nav.go_to_absolute(
+                robot_coords, 10000000000, distance_threshold, angle_threshold
+            )
             if blocking:
                 self.nav_result.wait()
         return "finished"
--- a/droidlet/lowlevel/hello_robot/remote/README.md
+++ b/droidlet/lowlevel/hello_robot/remote/README.md
@@ -20,6 +20,29 @@ popd
 
 ## 3. Start the services
 
+### Without ROS (i.e. directly using base API)
+
 ```bash
+# export LOCOBOT_IP=[your network ip]
+# example when using tailscale:
+# export LOCOBOT_IP=$(tailscale ip --4)
+
 ./launch.sh
+```
+
+### With ROS (i.e. better odometry using hector_slam, etc.)
+
+```bash
+# in fresh terminal
+deactivate
+./launch_ros.sh
+
+# in another terminal
+. droidlet/bin/activate
+
+# export LOCOBOT_IP=[your network ip]
+# example when using tailscale:
+# export LOCOBOT_IP=$(tailscale ip --4)
+
+./launch.sh --ros
 ```
--- a/droidlet/lowlevel/hello_robot/remote/README.md
+++ b/droidlet/lowlevel/hello_robot/remote/README.md
@@ -2,12 +2,14 @@
 
 On the Hello Robot Stretch, run the following from the directory where the readme exists
 
-## 1. Install venv and requirements
+## 1. Install venv and requirements and hector slam
 
 ```bash
 python3 -m venv droidlet
 . droidlet/bin/activate
 pip install -r requirements.txt
+
+sudo apt install ros-noetic-hector-slam -y
 ```
 
 ## 2. Install droidlet in develop mode
--- a/droidlet/lowlevel/hello_robot/remote/launch_ros.sh
+++ b/droidlet/lowlevel/hello_robot/remote/launch_ros.sh
@@ -1,0 +1,12 @@
+#!/bin/env bash
+set -e
+
+if $(which python | grep fairo >/dev/null); then
+    echo "droidlet virtualenv needs to be deactivated before running this script. Run 'deactivate' in the terminal before running this"
+    exit 0
+fi
+roslaunch stretch_laser_odom_base.launch &
+BGPID=$!
+trap 'echo "Killing $BGPID"; kill $BGPID; exit' INT
+sleep 5
+roslaunch stretch_hector_slam.launch
--- a/droidlet/lowlevel/hello_robot/remote/lidar_ros_driver.py
+++ b/droidlet/lowlevel/hello_robot/remote/lidar_ros_driver.py
@@ -1,0 +1,55 @@
+import math
+import sys
+import threading
+import time
+import traceback
+import rospy
+from sensor_msgs.msg import LaserScan
+
+
+class Lidar:
+    def __init__(self):
+        self._lock = threading.Lock()
+        self._thread = None
+        self.latest_scan = None
+
+    def start(self):
+        try:
+            rospy.init_node("droidlet_lidar_node")
+        except:
+            pass
+        self._thread = threading.Thread(target=self.lidar_loop, daemon=True)
+        self._thread.start()
+
+    def get_latest_scan(self):
+        with self._lock:
+            return self.latest_scan
+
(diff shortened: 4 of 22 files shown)


droidlet/lowlevel/hello_robot/remote/README.md
```diff
@@ -20,6 +20,29 @@ popd
...
+# export LOCOBOT_IP=$(tailscale ip --4)
+
 ./launch.sh
+```
+
+### With ROS (i.e. better odometry using hector_slam, etc.)
+
+```bash
+# in fresh terminal
+deactivate
```
cpaxton, line 37: Why do we need this?

droidlet/lowlevel/hello_robot/remote/README.md
```diff
@@ -20,6 +20,29 @@ popd
...
+./launch_ros.sh
```
cpaxton, line 38: extra steps for launching ROS - should we discuss? I think if its only one ROS launch file maybe it could go into submodule or MRP 

doesnt need to be a change for this PR though

droidlet/lowlevel/hello_robot/remote/launch_ros.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/env bash
+set -e
+
+if $(which python | grep fairo >/dev/null); then
+    echo "droidlet virtualenv needs to be deactivated before running this script. Run 'deactivate' in the terminal before running this"
```
cpaxton, line 5: will this cause issues?

droidlet/lowlevel/hello_robot/remote/lidar_ros_driver.py
```diff
@@ -0,0 +1,55 @@
+import math
+import sys
+import threading
+import time
+import traceback
+import rospy
+from sensor_msgs.msg import LaserScan
+
+
+class Lidar:
```
cpaxton, line 10: Should this share base class hierarchy with non-ROS version?

cpaxton: It would be nice to provide a commo interface

droidlet/lowlevel/hello_robot/remote/remote_hello_realsense.py
```diff
@@ -40,8 +39,14 @@
 class RemoteHelloRealsense(object):
     """Hello Robot interface"""
 
-    def __init__(self, bot):
+    def __init__(self, bot, use_ros=False):
         self.bot = bot
+        if use_ros:
+            self.use_ros = True
+            from droidlet.lowlevel.hello_robot.remote.lidar_ros_driver import Lidar
```
cpaxton, line 46: So they are the same - might be nice to have Lidar ABC?

droidlet/lowlevel/hello_robot/remote/remote_hello_robot.py
```diff
@@ -75,6 +75,9 @@ def _check_battery(self):
         val_in_range("CPU Temp", p.status["cpu_temp"], vmin=15, vmax=80)
         print(Style.RESET_ALL)
 
+    def pull_status(self):
+        self._robot.pull_status()
```
cpaxton, line 79: short comment to document function? not clear what this does to me

droidlet/lowlevel/hello_robot/remote/remote_hello_robot.py
```diff
@@ -233,10 +236,23 @@ def go_to_relative(self, xyt_position):
...
+            or robot.base.left_wheel.status["is_moving"]
+        )
+        right_wheel_moving = (
+            robot.base.right_wheel.status["is_moving_filtered"]
+            or robot.base.right_wheel.status["is_moving"]
+        )
+        is_moving = left_wheel_moving or right_wheel_moving
+        return is_moving
+
     def is_busy(self):
```
cpaxton, line 256: Should this include arm motion?

droidlet/lowlevel/hello_robot/remote/remote_hello_robot_ros.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+Pyro4.config.SERIALIZER = "pickle"
+Pyro4.config.SERIALIZERS_ACCEPTED.add("pickle")
+Pyro4.config.ITER_STREAMING = True
+
+
+# #####################################################
+@Pyro4.expose
+class RemoteHelloRobot(object):
```
cpaxton, line 28: Seems like a good amount of duplicate code here

cpaxton: maybe we should just create a base class for RemoteHelloRobot?

droidlet/lowlevel/hello_robot/remote/remote_hello_robot_ros.py
```diff
@@ -0,0 +1,232 @@
...
+        return self._robot.get_slam_pose()
+        # second Best (from lidar scanning)
+        # return self._robot.get_scan_matched_pose()
+        # Worst, from wheel encoder only
+        # return self._robot.get_odom()
+
+    def get_slam_pose(self):
+        return self._robot.get_slam_pose()
+
+    def pull_status(self):
```
cpaxton, line 105: not implemented - raise notimplementederror?

droidlet/lowlevel/hello_robot/remote/remote_hello_robot_ros.py
```diff
@@ -0,0 +1,232 @@
...
+    def get_tilt(self):
+        return self._robot.get_joint_state("head_tilt")
+
+    def set_pan(self, pan):
+        self._robot.send_command("joint_head_pan", pan)
+
+    def set_tilt(self, tilt):
+        self._robot.send_command("joint_head_tilt", tilt)
+
+    def reset_camera(self):
```
cpaxton, line 125: description - resets camera to home position?

droidlet/lowlevel/hello_robot/remote/stretch_ros_move_api.py
```diff
@@ -0,0 +1,192 @@
...
+from control_msgs.msg import FollowJointTrajectoryGoal
+from trajectory_msgs.msg import JointTrajectoryPoint
+from geometry_msgs.msg import PoseStamped, Pose2D, PoseWithCovarianceStamped
+from nav_msgs.msg import Odometry
+import hello_helpers.hello_misc as hm
+from tf.transformations import euler_from_quaternion
+import collections
+import numpy as np
+
+class MoveNode(hm.HelloNode):
```
cpaxton: I don't fully understand why we need this class - maybe it would make sense to add some comments?

droidlet/lowlevel/hello_robot/remote/stretch_ros_move_api.py
```diff
@@ -0,0 +1,192 @@
...
+                self._command = None                
+            if command is not None:
+                joint_name, increment = command
+                self._send_command(joint_name, increment)
+            rate.sleep()
+
+    def start(self):
+        hm.HelloNode.main(self, 'fairo_hello_proxy', 'fairo_hello_proxy', wait_for_first_pointcloud=False)
+        self._thread = threading.Thread(target=self.background_loop, daemon=True)
+        self._thread.start()
```
cpaxton, line 165: Remove dead code?

droidlet/lowlevel/hello_robot/remote/utils.py
```diff
@@ -238,20 +223,12 @@ def goto(
...
-            )
-            right_wheel_moving = (
-                robot.base.right_wheel.status["is_moving_filtered"]
-                or robot.base.right_wheel.status["is_moving"]
-            )
-            is_moving = left_wheel_moving or right_wheel_moving
+            is_moving = robot.is_base_moving()
 
     # second rotate by theta2
     theta_2 = np.sign(theta_2) * (abs(theta_2) % radians(360))
```
cpaxton, line 234: what are theta_1 vs theta_2?

droidlet/lowlevel/locobot/remote/slam_service.py
```diff
@@ -91,7 +92,7 @@ def add_obstacle(self, location, in_map=False):
         self.map_builder.add_obstacle(location)
 
     def update_map(self):
-        pcd = self.robot.get_current_pcd()[0]
+        pcd = safe_call(self.robot.get_current_pcd)[0]
```
cpaxton, line 95: What is safe_cal for? why is this necessary?

droidlet/lowlevel/hello_robot/remote/utils.py
```diff
@@ -196,7 +196,7 @@ def goto(
         robot.rotate_by(theta_1)
         robot.push_command()
         robot.pull_status()
-        time.sleep(0.1)
+        time.sleep(1)
```
cpaxton, line 199: any reason for these sleeps? should there be a generic "wait" for a status/command instead?

droidlet/lowlevel/hello_robot/hello_robot_mover.py
```diff
@@ -229,11 +229,11 @@ def move_relative(self, xyt_positions, blocking=True):
             xyt_positions = [xyt_positions]
         for xyt in xyt_positions:
             self.nav_result.wait()
-            self.nav_result = safe_call(self.nav.go_to_relative, xyt)
+            self.nav_result = safe_call(self.nav.go_to_relative, xyt, 10000000000, distance_threshold, angle_threshold)
```
cpaxton: strange hard coded value here

droidlet/lowlevel/hello_robot/hello_robot_mover.py
```diff
@@ -229,11 +229,11 @@ def move_relative(self, xyt_positions, blocking=True):
...
             if blocking:
                 self.nav_result.wait()
 
-    def move_absolute(self, xzt_positions, blocking=True):
+    def move_absolute(self, xzt_positions, blocking=True, distance_threshold=None, angle_threshold=None):
         """Command to execute a move to an absolute position.
 
         It receives positions in canonical world coordinates and converts them to pyrobot's coordinates
```
cpaxton, line 245: assuming None means that these are not used by default - but in that case how does it ever stop? does this have a timeout?

cpaxton: None might not be a good default if so

droidlet/lowlevel/hello_robot/hello_robot_mover.py
```diff
@@ -250,7 +250,7 @@ def move_absolute(self, xzt_positions, blocking=True):
             logging.info("Move absolute in canonical coordinates {}".format(xzt))
             self.nav_result.wait()
             robot_coords = base_canonical_coords_to_pyrobot_coords(xzt)
-            self.nav_result = self.nav.go_to_absolute(robot_coords)
+            self.nav_result = self.nav.go_to_absolute(robot_coords, 10000000000, distance_threshold, angle_threshold)
```
cpaxton: is 10000000 the timeout?

droidlet/lowlevel/locobot/remote/planning_service.py
```diff
@@ -76,27 +78,31 @@ def get_short_term_goal(self, robot_location, goal, step_size=25):
             target_goal = (stg_real[0], stg_real[1], rotation_angle)
         return target_goal
 
-    def goal_within_threshold(self, robot_location, goal):
-        distance = np.linalg.norm(np.array(robot_location[:2]) - np.array(goal[:2]))
-        # in metres. map_resolution is the resolution of the SLAM's 2D map, so the planner can't
-        # plan anything lower than this
-        threshold = (float(self.map_resolution) - 1e-10) / 100.0
+    def goal_within_threshold(self, robot_location, goal, threshold=None, angle_threshold=None):
```
cpaxton, line 82: description would be good - although this one seems pretty self explanatory

cpaxton: goal and location are the same type?

droidlet/lowlevel/hello_robot/remote/remote_hello_robot_ros.py
```diff
@@ -0,0 +1,271 @@
...
+        # at 90, the correction is +0.11526719 + -0.091491526943
+        # linear interpolate the correction of 0.023775
+        interp_correction = 0.11526719 * abs(head_tilt) / radians(90)
+        # print('interp_correction', interp_correction)
+
+        camera_transform[2, 3] += -0.091491526943 + interp_correction
+
+        return camera_transform
+
+    def get_base_state(self):
```
theophilee, line 95: The `yaw` is in `[-3.14, 3.14]` when using Hector SLAM and in `[0, 6.28]` when not, maybe we could make it consistent?

droidlet/lowlevel/hello_robot/remote/README.md
soumith: this is to make sure we aren't in any kind of virtualenv / pipenv. The ROS launch commands need to use system python as-is.

droidlet/lowlevel/hello_robot/remote/README.md
soumith: Will move everything to MRP in an upcoming PR

droidlet/lowlevel/hello_robot/remote/launch_ros.sh
soumith: not really. it's a check to make sure issues aren't caused by accidentally forgetting to do `deactivate` or having a virtualenv active

droidlet/lowlevel/hello_robot/remote/remote_hello_realsense.py
soumith: done

droidlet/lowlevel/hello_robot/remote/remote_hello_robot_ros.py
soumith: This is intentionally intended to be empty, because we want to match the API between this and the non-ROS class

soumith: we dont need to do a pull_status with the ROS driver

droidlet/lowlevel/hello_robot/remote/remote_hello_robot_ros.py
soumith: done

droidlet/lowlevel/hello_robot/remote/stretch_ros_move_api.py
soumith: done. yes it's the Move API but via ROS.
Once the robot is driven via the ROS stuff (i.e. to use hector_slam), then we cant use the move API directly anymore and are forced to control the robot via ROS as well

droidlet/lowlevel/hello_robot/remote/stretch_ros_move_api.py
soumith: done

droidlet/lowlevel/hello_robot/remote/utils.py
soumith: they're described in the comments in the file.

droidlet/lowlevel/locobot/remote/slam_service.py
soumith: needed for making pyro4 fail-fast in case of a downstream service error (instead of waiting to timeout).
All the safe_call business will go away once we move to zerorpc.

droidlet/lowlevel/hello_robot/hello_robot_mover.py
soumith: removed

droidlet/lowlevel/hello_robot/hello_robot_mover.py
```diff
@@ -217,7 +217,7 @@ def reset_camera(self):
         """reset the camera to 0 pan and tilt."""
         return self.bot.reset()
 
-    def move_relative(self, xyt_positions, blocking=True):
+    def move_relative(self, xyt_positions, blocking=True, distance_threshold=None, angle_threshold=None):
         """Command to execute a relative move.
 
         Args:
```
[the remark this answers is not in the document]
soumith: done

droidlet/lowlevel/hello_robot/hello_robot_mover.py
soumith: None will make it default to the planner's default values, which are dynamic based on the map resolution.
For example, the default planner's distance threshold is:
```
threshold = (float(self.map_resolution) - 1e-10) / 100.0
```

soumith: this makes sense, because the planner cant plan a path with a resolution lower than it's map resolution anyways

droidlet/lowlevel/hello_robot/hello_robot_mover.py
soumith: removed it

droidlet/lowlevel/locobot/remote/planning_service.py
soumith: goal and location are of same type, indeed.

droidlet/lowlevel/hello_robot/remote/remote_hello_robot_ros.py
soumith: done. added a `math.pi` to the yaw in this case, so that it's consistent to `[0, 6.28]`.

droidlet/lowlevel/hello_robot/remote/remote_hello_robot_ros.py
```diff
@@ -197,13 +205,22 @@ def go_to_relative(self, xyt_position):
...
+                print(e)
+                print(traceback.format_exc())
+                self._done = True
+                raise e
         return status
 
     def is_moving(self):
         return not self._done
 
+    def is_busy(self):
```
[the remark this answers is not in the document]
soumith: yes, when we add in the API to move the Arm

droidlet/lowlevel/hello_robot/remote/utils.py
soumith: there isn't a generic `wait` available in the base API for stretch.
But I spoke to the Hello folks a few days back and there's a better way to handle this than the sleep.
Will implement in a follow-up PR

droidlet/lowlevel/hello_robot/remote/remote_hello_robot_ros.py
soumith: will open an issue for it.

droidlet/lowlevel/hello_robot/remote/remote_hello_robot.py
soumith: yes, but after we implement the arm capabilities

cpaxton (approved):

PR #1162, opened 2022-05-24
explore on slurm

anuragprat1k:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* (to be filled)

### DIFF
--- a/tools/active_vision/explore/explore.py
+++ b/tools/active_vision/explore/explore.py
@@ -1,0 +1,75 @@
+import argparse
+import submitit
+from pathlib import Path
+import os
+import subprocess
+import json
+import logging
+from typing import List
+
+class explore_job:
+    def __init__(self):
+        pass 
+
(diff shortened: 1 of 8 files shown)


tools/active_vision/explore/explore.py
```diff
@@ -0,0 +1,86 @@
+import argparse
```
aszlam, line 1: could this file go in fairo/droidlet/tools instead of fairo/tools?

aszlam (approved):

tools/active_vision/explore/explore.py
anuragprat1k: Moved these to fairo/droidlet/tools a few commits later in this stack (https://github.com/facebookresearch/fairo/pull/1166).

soumith (approved):

PR #1170, opened 2022-05-24
label prop

anuragprat1k:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* (to be filled)

### DIFF
--- a/droidlet/perception/robot/handlers/label_propagate.py
+++ b/droidlet/perception/robot/handlers/label_propagate.py
@@ -3,25 +3,26 @@ Copyright (c) Facebook, Inc. and its affiliates.
 """
 from .core import AbstractHandler
 import numpy as np
-import os
-import cv2
-import json
-import glob
 from droidlet.lowlevel.robot_mover_utils import transform_pose
 from numba import njit
-from math import ceil, floor
+from math import ceil, floor, isnan
+from collections import deque, defaultdict
+import random
+import time
 
-# Values for locobot in habitat.
+# Values for locobot in habitat. 
 # TODO: generalize this for all robots
 fx, fy = 256, 256
 cx, cy = 256, 256
-intrinsic_mat = np.array([[fx, 0.0, cx], [0.0, fy, cy], [0.0, 0.0, 1.0]])
+intrinsic_mat = np.array([[  fx, 0., cx],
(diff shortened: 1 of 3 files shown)


droidlet/perception/robot/handlers/label_propagate.py
```diff
@@ -76,41 +145,56 @@ def __call__(
...
+        valid_z = src_pts_in_cur_cam[:, 2] > 0
+
+        # Filter based on current depth.
+        cur_depth = (cur_depth.astype(np.float32) / 1000.0).reshape(-1)
+        cur_pts_in_cur_cam = np.multiply(uv_one_in_cam, cur_depth).T
 
-        return get_annot(height, width, pts_in_cur_img, src_label)
+        # conver pts in current camera frame into 2D pix values
+        src_pts_in_cur_img = np.matmul(intrinsic_mat, src_pts_in_cur_cam.T).T
+        src_pts_in_cur_img /= src_pts_in_cur_img[:, 2].reshape([-1, 1])
```
soumith, line 190: i think this is a bug. you shouldn't divide `src_pts_in_cur_img[:, 2] / src_pts_in_cur_img[:, 2]` right?
You should only do: `src_pts_in_cur_img[:, 0:2] / src_pts_in_cur_img[:, 2]`

anuragprat1k: `src_pts_in_cur_img[:, 2]` needs to be reshaped for the division. So this does that by `src_pts_in_cur_img[:, 2].reshape([-1, 1])` which is essentially a `num_row * 1` array consisting of the last column of `src_pts_in_cur_img` and it is then broadcasted to divide each column of `src_pts_in_cur_img`

soumith: yes, but you are dividing `z / z`, which will make all `z` values to be `1.0`.

The correct transform is what I wrote in `uncompute_pcd`, which is only `[:, 0] / [:, 2]` and `[:, 1] / [:, 2]` need to be done.

anuragprat1k: That was intended - wanted to keep the homogenous image coordinates in the format `(x,y,1)`.

Issue #1192, opened 2022-06-01
Breakage from Querying Selection of all Reference Objects without WHERE Clause


aqibisbiqa:
## Type of Issue

Select the type of issue:
- [x] Bug report (to report a bug)
- [ ] Feature request (to request an additional feature)
- [ ] Tracker (I am just using this as a tracker)
- [ ] Refactor request
- [ ] Documentation Ask

## Description

Title explains issue. Motivation due to this error:

```
TypeError: cannot unpack non-iterable NoneType object
Traceback (most recent call last):
  File "agents/craftassist/craftassist_agent.py", line 568, in <module>
    sa.start()
  File "/private/home/aqib/fairo/agents/core.py", line 23, in start
    self.handle_exception(e)
  File "/private/home/aqib/fairo/agents/droidlet_agent.py", line 296, in handle_exception
    raise e
  File "/private/home/aqib/fairo/agents/core.py", line 21, in start
    self.step()
  File "/private/home/aqib/fairo/agents/droidlet_agent.py", line 301, in step
    super().step()
  File "/private/home/aqib/fairo/agents/core.py", line 26, in step
    self.perceive()
  File "agents/craftassist/craftassist_agent.py", line 298, in perceive
    self.draw_map_to_dashboard()
  File "agents/craftassist/craftassist_agent.py", line 414, in draw_map_to_dashboard
    detections_for_map = self.get_detected_objects_for_map()
  File "agents/craftassist/craftassist_agent.py", line 399, in get_detected_objects_for_map
    search_res = self.memory.basic_search("SELECT MEMORY FROM ReferenceObject")
  File "/private/home/aqib/fairo/droidlet/memory/sql_memory.py", line 369, in basic_search
    return self.searcher.search(self, query=query, get_all=get_all)
  File "/private/home/aqib/fairo/droidlet/memory/memory_filters.py", line 516, in search
    vals = self.handle_output(agent_memory, query, memids, get_all)
  File "/private/home/aqib/fairo/droidlet/memory/memory_filters.py", line 451, in handle_output
    return [agent_memory.get_mem_by_id(m) for m in memids]
  File "/private/home/aqib/fairo/droidlet/memory/memory_filters.py", line 451, in <listcomp>
    return [agent_memory.get_mem_by_id(m) for m in memids]
  File "/private/home/aqib/fairo/droidlet/memory/sql_memory.py", line 305, in get_mem_by_id
[...]

PR #1200, opened 2022-06-05
Fix nowhere bug

aszlam:
# Description

Corrects basic search queries with no WHERE clause.   Previous bug was due to archives being returned along with active memories despite not having an entry in the relevant table.  along the way this PR defaults agent memory to build a SelfNode (not doing so was breaking the tests).

Fixes # (issue)
https://github.com/facebookresearch/fairo/issues/1192

## Type of change

- [x] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

# Checklist:

### DIFF
--- a/agents/craftassist/craftassist_agent.py
+++ b/agents/craftassist/craftassist_agent.py
@@ -18,7 +18,7 @@ from copy import deepcopy
 # also used as a standalone script and invoked via `python craftassist_agent.py`
 from droidlet.interpreter.craftassist import default_behaviors, inventory, dance
 from droidlet.memory.craftassist import mc_memory
-from droidlet.memory.memory_nodes import ChatNode
+from droidlet.memory.memory_nodes import ChatNode, SelfNode
 from droidlet.shared_data_struct import rotation
 from droidlet.lowlevel.minecraft.craftassist_mover import (
     CraftassistMover,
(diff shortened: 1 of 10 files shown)


droidlet/memory/memory_nodes.py
```diff
@@ -784,6 +785,36 @@ class SelfNode(PlayerNode):
...
+
+        """
+        memid = memid or cls.new(memory)
+        if player_struct is None:
+            eid, name, x, y, z, pitch, yaw = None, None, None, None, None, None, None
+        else:
+            eid, name = player_struct.entityId, player_struct.name
+            x, y, z = player_struct.pos
+            yaw, pitch = player_struct.look
+        cmd = "INSERT INTO ReferenceObjects(uuid, eid, name, x, y, z, pitch, yaw, ref_type) VALUES (?,?,?,?,?,?,?,?,?)"
```
kavyasrinet, line 803: to be clear, earlier we were doing this for the agents here: https://github.com/facebookresearch/fairo/blob/main/droidlet/memory/sql_memory.py#L140 
which was equivalent to doing this anyway, right ?

agents/craftassist/craftassist_agent.py
```diff
@@ -499,7 +500,7 @@ def add_self_memory_node(self):
             p = self.get_player()
         except:  # this is for test/test_agent
             return
-        PlayerNode.create(self.memory, p, memid=self.memory.self_memid)
```
kavyasrinet, line 502: Quick question - since we already create a PlayerNode/SelfNode in `sql_memory.py` why do we need to also create one here ?

droidlet/memory/memory_nodes.py
aszlam: it didn't set it as a ReferenceObject; that was being done in the agent.

agents/craftassist/craftassist_agent.py
aszlam: we don't- instead we just do the update now to make sure the agent's position and look are stored (they are Null when the SelfNode is created in memory because the agent isn't fully instantiated yet)

kavyasrinet (approved):

PR #1218, opened 2022-06-10
Add JointVelocityControl

igormolybogFB:
Summary:

Test Plan:
python ./polymetis/tests/scripts/7_joint_velocities.py 

Reviewers:

Subscribers:

Tasks:
T122819435

Tags:

# Description

Please include a summary of the change and link to an issue this PR fixes. Please also include relevant context. List any downstream components that might be affected by this change.

Fixes # (issue)

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

# Testing

# Checklist:

### DIFF
--- a/polymetis/polymetis/python/polymetis/robot_interface.py
+++ b/polymetis/polymetis/python/polymetis/robot_interface.py
@@ -588,6 +588,38 @@ class RobotInterface(BaseRobotInterface):
 
         return update_idx
 
+
+    def start_joint_velocity(self, joint_vel_desired, Kq=None, Kqd=None, **kwargs):
+        """Starts joint position control mode.
+        Runs an non-blocking joint velocity controller.
+        The desired joint velocities can be updated using `update_desired_joint_velocities`
+        """
+        torch_policy = toco.policies.JointVelocityControl(
+            joint_vel_desired=joint_vel_desired,
+            Kp=self.Kq_default if Kq is None else Kq,
+            Kd=self.Kqd_default if Kqd is None else Kqd,
+            robot_model=self.robot_model,
+            ignore_gravity=self.use_grav_comp,
+        )
+
+        return self.send_torch_policy(torch_policy=torch_policy, blocking=False)
+
+
+    def update_desired_joint_velocities(self, velocities: torch.Tensor):
+        """Update the desired joint positions used by the joint position control mode.
+        Requires starting a joint impedance controller with `start_joint_impedance` beforehand.
+        """
+        try:
+            update_idx = self.update_current_policy({"joint_vel_desired": velocities})
+        except grpc.RpcError as e:
+            log.error(
+                "Unable to update desired joint velocities. Use 'start_joint_velocities' to start a joint velocities controller."
+            )
+            raise e
+
+        return update_idx
+
+
     """
     PyRobot backward compatibility methods
     """
--- a/polymetis/polymetis/python/polymetis/robot_interface.py
+++ b/polymetis/polymetis/python/polymetis/robot_interface.py
@@ -589,7 +589,7 @@ class RobotInterface(BaseRobotInterface):
         return update_idx
 
 
-    def start_joint_velocity_control(self, joint_vel_desired, Kq=None, Kqd=None, **kwargs):
+    def start_joint_velocity_control(self, joint_vel_desired, hz=None, Kq=None, Kqd=None, **kwargs):
         """Starts joint velocity control mode.
         Runs a non-blocking joint velocity controller.
         The desired joint velocities can be updated using `update_desired_joint_velocities`
@@ -599,6 +599,7 @@ class RobotInterface(BaseRobotInterface):
             Kp=self.Kq_default if Kq is None else Kq,
             Kd=self.Kqd_default if Kqd is None else Kqd,
             robot_model=self.robot_model,
+            hz=self.metadata.hz if hz is None else hz,
             ignore_gravity=self.use_grav_comp,
         )
 
--- a/polymetis/polymetis/python/polymetis/robot_interface.py
+++ b/polymetis/polymetis/python/polymetis/robot_interface.py
@@ -589,9 +589,9 @@ class RobotInterface(BaseRobotInterface):
         return update_idx
 
 
-    def start_joint_velocity(self, joint_vel_desired, Kq=None, Kqd=None, **kwargs):
-        """Starts joint position control mode.
-        Runs an non-blocking joint velocity controller.
+    def start_joint_velocity_control(self, joint_vel_desired, Kq=None, Kqd=None, **kwargs):
+        """Starts joint velocity control mode.
+        Runs a non-blocking joint velocity controller.
         The desired joint velocities can be updated using `update_desired_joint_velocities`
         """
         torch_policy = toco.policies.JointVelocityControl(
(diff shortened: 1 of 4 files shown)


polymetis/polymetis/python/polymetis/robot_interface.py
```diff
@@ -588,6 +588,38 @@ def update_desired_ee_pose(
 
         return update_idx
 
+
+    def start_joint_velocity(self, joint_vel_desired, Kq=None, Kqd=None, **kwargs):
```
exhaustin, line 592: I would rename to `start_joint_velocity_control` to make it less confusing.
(`impedance` is less ambiguous in comparison for the other methods)

polymetis/polymetis/python/polymetis/robot_interface.py
```diff
@@ -588,6 +588,38 @@ def update_desired_ee_pose(
...
+        """Starts joint position control mode.
```
exhaustin, line 593:
```suggestion
        """Starts joint velocity control mode.
```

polymetis/polymetis/python/polymetis/robot_interface.py
```diff
@@ -588,6 +588,38 @@ def update_desired_ee_pose(
...
+            robot_model=self.robot_model,
+            ignore_gravity=self.use_grav_comp,
+        )
+
+        return self.send_torch_policy(torch_policy=torch_policy, blocking=False)
+
+
+    def update_desired_joint_velocities(self, velocities: torch.Tensor):
+        """Update the desired joint positions used by the joint position control mode.
+        Requires starting a joint impedance controller with `start_joint_impedance` beforehand.
```
exhaustin, line 610: Remember to also update this docstring before merging

polymetis/polymetis/python/torchcontrol/policies/velocity.py
```diff
@@ -0,0 +1,85 @@
...
+        # current_timestamp = state_dict["timestamp"]
+        if self.timestep_initialized:
+            dt = torch.dot((state_dict["timestamp"]-self.previous_timestamp).to(torch.float), self.timestep_to_seconds_transformation)
+        else:
+            # In the first timestep, dt is impossible to estimate so best guess is 0.0
+            dt = torch.tensor(0.0)
+            self.timestep_initialized = True
+        self.previous_timestamp = torch.clone(state_dict["timestamp"])
+
+        joint_pos_desired = joint_pos_current + torch.mul(self.joint_vel_desired, dt)
```
exhaustin, line 71: You might want to use `joint_pos_desired_prev` instead of using `joint_pos_current`
Or even better, simply maintain a `joint_pos_desired` and update every timestep by `self.joint_pos_desired += torch.mul(self.joint_vel_desired, dt)`
The reason being the original formulation of the PI controller is to integrate `(v_desired - v_current) * dt`, which is the integral of `v_desired * dt` (**desired position resulting from all desired velocities**) - the integral of `v_current * dt` (current position).

exhaustin (changes requested):
Generally looks good and code looks really clean!
The most significant issue is the part of the control logic that is unfaithful to the original formulation, which might be what's causing the inaccuracies.

I'm also thinking that I might be wrong in our previous discussions and providing hz from the user side would be a better idea. The reason being 1) there might be errors at which the timestamps were measured, and 2) cleaner and simpler controller logic.

polymetis/polymetis/python/torchcontrol/policies/velocity.py
```diff
@@ -21,6 +21,7 @@ def __init__(
         Kp,
         Kd,
         robot_model: torch.nn.Module,
+        hz=torch.tensor(0, dtype=torch.int32),
```
exhaustin, line 24: Since the constructor is being run on the python side, there is no need to constrain arguments to be torch tensors.
I would designate `hz` as a required int argument and remove the part where it is designated on the fly.

Also, since `self.hz` is not of type tensor, remember to annotate the attribute as explained in the [TorchScript docs](https://pytorch.org/docs/stable/jit.html#torchscript-classes).

polymetis/polymetis/python/polymetis/robot_interface.py
```diff
@@ -588,6 +588,38 @@ def update_desired_ee_pose(
...
+        Runs a non-blocking joint velocity controller.
+        The desired joint velocities can be updated using `update_desired_joint_velocities`
+        """
+        torch_policy = toco.policies.JointVelocityControl(
+            joint_vel_desired=joint_vel_desired,
+            Kp=self.Kq_default if Kq is None else Kq,
+            Kd=self.Kqd_default if Kqd is None else Kqd,
+            robot_model=self.robot_model,
+            ignore_gravity=self.use_grav_comp,
+        )
```
exhaustin, line 603: Add argument `hz=self.metadata.hz` here.

polymetis/polymetis/python/torchcontrol/policies/velocity.py
```diff
@@ -0,0 +1,87 @@
...
+
+import torchcontrol as toco
+from torchcontrol.utils import to_tensor
+
+
+class JointVelocityControl(toco.PolicyModule):
+    """
+    Velocity control in joint space.
+    """
+    hz: int
```
exhaustin, line 17: Also add:
```
dt: float
is_initialized: bool
```

Also see below comment for dropping the `hz` member variable.

polymetis/tests/scripts/7_joint_velocities.py
```diff
@@ -0,0 +1,37 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+import time
+
+import torch
+
+from polymetis import RobotInterface
+from utils import check_episode_log
```
exhaustin, line 10: No longer needed.

polymetis/tests/scripts/7_joint_velocities.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def test_new_joint_vel(state_log, joint_vel_desired):
+    joint_vel = state_log[-1].joint_velocities
```
exhaustin, line 14: Since the velocity controller aims to maintain the desired velocity for the duration of the controller, you should check all velocities within the state log instead of just the last one. The first few steps can be left out as it's normal for the controller to need time to stabilize.

polymetis/tests/scripts/7_joint_velocities.py
```diff
@@ -0,0 +1,37 @@
...
+
+    # Joint velocity control
+    print("=== RobotInterface.start_joint_velocity_control ===")
+
+    joint_vel_desired = torch.Tensor([0.1, -0.1, -0.1, 0.0, 0.0, 0.1, 0.1])
+
+    robot.start_joint_velocity_control(joint_vel_desired)
+    time.sleep(3)
+    state_log = robot.terminate_current_policy()
+    joint_vel = test_new_joint_vel(state_log, joint_vel_desired)
```
exhaustin, line 37: I think this is a good test in general, but it does not cover the new method `robot.update_desired_joint_velocities`.
Can you include this in the test? 
Off the top of my head maybe update it once at 1.5 seconds and check the joint velocities of the two halves separately, but feel free to do it however you deem proper.

exhaustin (changes requested):
Almost there, just a few more minor nits :)

I will also be testing this on a real robot today, will let you know how it goes.

exhaustin (approved):
Looks great, thank you!

PR #1207, opened 2022-06-07
Enable memory_2d map for Craftassist Agent

aqibisbiqa:
# Description

(Fresh version of [earlier PR](https://github.com/facebookresearch/fairo/pull/1180))
The map that displays in the in the memory_2d pane for the Locobot agent was not previously implemented for the Craftassist agent. With this PR, we can now see the agent, obstacles from the obstacle map, and other detected objects as points plotted on the map pane.

Fixes # (issue)

## Type of change

Please check the options that are relevant.

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Proposes a change (non-breaking change that isn't necessarily a bug)
- [ ] Refactor
- [x] New feature (non-breaking change that adds a new functionality)
- [ ] Breaking change (fix or feature that would break some existing functionality downstream)
- [ ] This is a unit test
- [ ] Documentation only change
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

- [x] I want a thorough review of the implementation.
- [ ] I want a high level review. 
- [ ] I want a deep design review.

## Before and After

Before:
Screen Shot 2022-05-20 at 4 26 50 PM

After:
Screen Shot 2022-06-07 at 2 49 58 PM

# Testing

You can see a map :) And the agent program + dashboard doesn't break when rendering the map after a multitude of test commands. This diff won't affect any of the previously implemented unit tests, and it seems to pass all of the existing ones.

Oftentimes you'll only see the bot marker on the map, you can populate the agent's perceived obstacle map by issuing a command. For example, to produce the 'After' picture, run your cuberite instance with command `python droidlet/lowlevel/minecraft/cuberite_process.py --config really_flat_world` and issue the command `move three steps left`.

# Checklist:

- [x] I have performed manual end-to-end testing of the feature in my environment.
- [x] I have added Docstrings and comments to the code.
- [x] I have made changes to existing documentation where needed.
- [ ] I have added tests that show that the PR is functional.
- [ ] New and existing unit tests pass locally with my changes.
- [x] I have added relevant collaborators to review the PR before merge.
- [ ] [Polymetis only] I ran on hardware (1) all scripts in `tests/scripts`, (2) asv benchmarks.

PR #1220, opened 2022-06-10
Memory Map performance fixes

aqibisbiqa:
# Description

TLDR;
* Map disabled by default
* Can now enable/disable the map in frontend with a simple switch in Settings pane, disabling massively saves rendering costs
* Disabling in frontend sends signal to agent to not send map updates
* Altered options to enable/disable map for backend
* Can specify how many ticks Craftassist agent sends map updates (default 20 ticks = 1 second)

This PR primarily addresses the drop in performance of the Craftassist agent introduced in my [#1207](https://github.com/facebookresearch/fairo/pull/1207). Essentially, all the socket-io emits from the agent to update the dashboard clog the connection between the front and back end. 

BACKEND: Disabling the map sends a signal to the agent to not send (or calculate) any map updates at all. For craftassist, if the map is enabled, the agent will now only send updates every N ticks (that you can specify). The default is every 20 ticks (1 second). Some of the argument parsing was also modified:
* `python craftassist_agent.py` will run the agent, but the agent won't send any map updates
* `python craftassist_agent.py --draw_map --map_update_ticks 60` will run the agent and send map updates every 60 ticks
* `python locobot_agent.py --draw_map --map_data "memory"` will run the agent and draw the map from memory

FRONTEND: one can disable the memory map entirely with a switch in the settings pane. By default the map is disabled (for both Robot and Craftassist), so you need to click the switch to see anything on the map. If the map doesn't load immediately after clicking the switch, it is possible that the agent hasn't entirely properly loaded. If the map still doesn't load, try clicking the map pane and it should appear. The performance savings here come from the new `enable_map` state variable, which, if false, does an early return of the memory map render and skips all the regular map operations.

Note that to enable the map, you need to do TWO things: use the switch introduced in this PR, and specify the `--draw_map` option. Also, the switch won't even appear if you haven't specified the `--draw_map` option.

Fixes # (issue)

## Type of change

- [x] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

You can try running craftassist from `main` right now and see how slow the agent runs when issuing a command. Then, try running from this branch (`aqib/dash_perf_opt`) and you can see the increase in performance when the map is disabled.

**Before:**
Screen Shot 2022-06-13 at 8 47 39 AM

**After:**

Initial View
Screen Shot 2022-06-13 at 8 42 59 AM

After switching on map and clicking on pane
Screen Shot 2022-06-13 at 8 43 13 AM

After switching off map and clicking on pane
Screen Shot 2022-06-13 at 8 43 21 AM

# Testing

See above

# Checklist:


agents/argument_parser.py
```diff
@@ -100,8 +100,12 @@ def add_mc_parser(self):
         )
         mc_parser.add_argument(
             "--draw_map", 
-            default="memory",
```
soumith, line 103: locobot_agent.py also uses this arg, so that has to be updated as well: https://github.com/facebookresearch/fairo/blob/2e00b7b42519e27cfaebd25cec3c6327b79759df/agents/locobot/locobot_agent.py#L268-L273

soumith: maybe a better way would be to keep this option as-is but rename it `map_data`.
Then, also have a flag `draw_map` that either enables it or not.

soumith (changes requested):

agents/argument_parser.py
aqibisbiqa: > maybe a better way would be to keep this option as-is but rename it `map_data`. Then, also have a flag `draw_map` that either enables it or not.

Ah ok, that makes sense to break it down. Thanks Soumith!

soumith (approved):

kavyasrinet (approved):
Look good after the fixes.

PR #1237, opened 2022-06-17
[Polymetis] New Polysim API

exhaustin:
# Description

New API of Polysim that allows simulation in user code to easily connect to Polymetis.
Also enables simulation to connect to multiple separate control servers (e.g. arm control server & gripper control server)

[Design doc](https://docs.google.com/document/d/1PR0iKdCC13Z45WXv5asbeGzCwmVhlx2413MTZx4NDiM/edit#)

### Changes
- Allow Polymetis servers to launch without a client by specifying `robot_client=none` or `gripper=none`
- Add new Polysim API for connecting sim to servers
- Add metadata for grippers
- Add new experimental bullet manipulator env that supports control of both arm & gripper
- Add minimal double integrator example

### Instructions to run new Bullet manipulator with a gripper

1. Launch both servers without a client:
```
launch_robot.py robot_client=none
launch_gripper.py gripper=none
```
2. Launch the new simulation client which is capable of connecting to both the robot server and the gripper server:
```
python -m polysim.envs.experimental.bullet_manipulator gui=<true | false>
```

4. Connect to servers with `RobotInterface` and `GripperInterface` in user code to control both the arm and gripper in the simulation.

### Todos
- [ ] Finalize sim API
- [ ] Debug minimal example
- [ ] Add tests (after initial design review)
- [ ] Update documentation (after initial design review)

### Design problems

Metadata specification
- User-facing Polysim API should allow the user to connect to server with an easy-to-setup metadata
- It's ideal for Polymetis to provide a set of well-tuned metadata to go with predefined envs (and provide accurate sim2real), which requires complicated metadata config

Solution 1: Make metadata object initialize-able from select arguments or a full config file?
Solution 2: Make most fields of metadata optional and serialize into a single str or byte protobuf field? (experiemented, but reverted in this [commit](https://github.com/facebookresearch/fairo/pull/1237/commits/d9df607f150df119f393c2a7e4f1ac0fef4dd284))

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

# Testing

# Checklist:


exhaustin:
Currently missing tests and documentation. Due to the large structural changes, would like a deep design review before finalizing the PR.

polymetis/polymetis/python/polysim/envs/experimental/bullet_manipulator.py
```diff
@@ -0,0 +1,164 @@
...
+
+        urdf_path = get_full_path_to_urdf(self.cfg.robot_description_path)
+        self.robot_id = self.sim.loadURDF(
+            urdf_path,
+            basePosition=[0.0, 0.0, 0.0],
+            useFixedBase=True,
+            flags=pybullet.URDF_USE_INERTIA_FROM_FILE,
+        )
+
+        for i in range(7):
```
lshamis, line 41: Are we tying the manipulation simulation in bullet to 7-dof arms?

polymetis/polymetis/python/polysim/envs/experimental/bullet_manipulator.py
```diff
@@ -0,0 +1,164 @@
...
+        self.arm_state.motor_torques_external[:] = np.zeros(7)
+
+        self.arm_state.prev_command_successful = True
+        self.arm_state.error_code = 0
+
+        self.gripper_state = polymetis_pb2.GripperState()
+
+    def get_arm_state(self) -> polymetis_pb2.RobotState:
+        # Timestamp
+        self.arm_state.timestamp.GetCurrentTime()
```
lshamis, line 63: Is this the right timestamp?

polymetis/polymetis/python/polysim/envs/experimental/bullet_manipulator.py
```diff
@@ -0,0 +1,164 @@
...
+        joint_cur_states = self.sim.getJointStates(
+            self.robot_id, self.cfg.controlled_joints
+        )
+        self.arm_state.joint_positions[:] = [joint_cur_states[i][0] for i in range(7)]
+        self.arm_state.joint_velocities[:] = [joint_cur_states[i][1] for i in range(7)]
+
+        return self.arm_state
+
+    def get_gripper_state(self) -> polymetis_pb2.GripperState:
+        joint_cur_states = self.sim.getJointStates(
```
lshamis, line 75: Comparing this to get_arm_state, does gripper_state not need a timestamp?

polymetis/polymetis/python/polysim/envs/experimental/bullet_manipulator.py
```diff
@@ -0,0 +1,164 @@
...
+        )
+
+        # Populate torques in state
+        self.arm_state.prev_joint_torques_computed[:] = commanded_torques
+        self.arm_state.prev_joint_torques_computed_safened[:] = commanded_torques
+        self.arm_state.motor_torques_measured[:] = applied_torques
+        self.arm_state.motor_torques_external[:] = np.zeros_like(applied_torques)
+
+        self.arm_state.prev_command_successful = True
+        self.arm_state.error_code = 0
```
lshamis, line 121: When is error not 0?

polymetis/polymetis/python/polysim/sim_interface.py
```diff
@@ -0,0 +1,156 @@
...
+from omegaconf import OmegaConf
+from omegaconf.dictconfig import DictConfig
+
+import torchcontrol as toco
+from torchcontrol.policies.default_controller import DefaultController
+import polymetis_pb2
+import polymetis_pb2_grpc
+import polymetis
+from polymetis.utils import Spinner
+from polymetis.robot_client.abstract_robot_client import (
```
lshamis, line 29: Please trim down the imports, I don't think many are used.

polymetis/polymetis/python/polysim/sim_interface.py
```diff
@@ -0,0 +1,156 @@
...
+        self.fast_forward = fast_forward
+
+        self.control_items = []
+        self.step_callback = None
+
+    @staticmethod
+    def _serialize_controller(controller):
+        buffer = io.BytesIO()
+        torch.jit.save(torch.jit.script(controller), buffer)
+        buffer.seek(0)
```
lshamis, line 68: Doesn't really matter, but you can return `buffer.getvalue()` without seeking

polymetis/polymetis/python/polysim/sim_interface.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def register_gripper_control(
+        self,
+        server_address: str,
+        state_callback: Callable,
+        action_callback: Callable,
+        max_width: float,
+    ):
+        # Construct metadata
+        metadata = polymetis_pb2.GripperMetadata()
```
lshamis, line 122: I think you created a helper function just for this:
```py
metadata = self._construct_metadata(
    polymetis_pb2.GripperMetadata(),
    {"max_width": max_width})
```

polymetis/polymetis/python/scripts/launch_gripper.py
```diff
@@ -20,8 +21,12 @@ def main(cfg):
...
     else:
         # Run client
-        gripper_client = hydra.utils.instantiate(cfg.gripper)
-        gripper_client.run()
+        if cfg.gripper:
+            gripper_client = hydra.utils.instantiate(cfg.gripper)
+            gripper_client.run()
+
+        else:
+            signal.pause()
```
lshamis, line 29: If someone launches gripper and the config has no gripper, why run the server and block the child process on signal? Which signal are you expecting?

Depending on how this is intended to be used, it might be better to check before the fork and die early.

polymetis/polymetis/python/scripts/launch_robot.py
```diff
@@ -79,15 +79,19 @@ def cleanup():
...
+            time.sleep(0.1)
+            if time.time() - t0 > cfg.timeout:
+                raise ConnectionError("Robot client: Unable to locate server.")
+
+        log.info(f"Starting robot client...")
+        client = hydra.utils.instantiate(cfg.robot_client)
+        client.run()
+
+    else:
+        signal.pause()
```
lshamis, line 94: Same question as the gripper.

1heart (commented):
I understand this is a draft PR meant to demo the design, so I won't comment too much on details of the implementation.

Broadly speaking, I think the callback-based `register_*_control` design makes sense to me. The points I'd like to emphasize are:

1. User experience of adding a simulator. The set of metadata required is pretty large; hopefully much of this can be generated from the urdf (e.g. dof) or have sane defaults (server_address, possibly the gains).
    1. I'm not sure if we're moving immediately to an in-process controller manager or not, but ideally someone adding a simulation doesn't need to specify a server address. This leads to the next point:
3. User experience of launching a simulator. Ideally we don't need to launch a bunch of controller manager servers before launching their simulation. Perhaps this is something mrp will take care of (in which case we add an example `msetup.py`), or if we move to in-process controller manager that's even better.
4. Distinction between arm vs. gripper control (i.e. RobotInterface and GripperInterface(. In my head, the current distinction is between torque-controlled many-DoF robot vs. basic 2-finger gripper, which doesn't feel super satisfactory to me (GripperInterface in particular). Is there a cleaner, more abstract version of GripperInterface which allows us to extend it somewhere?

Happy to discuss this over VC. I think the main point to settle on is a minimal, extensible `*_control` abstraction, such that as we add different robots / end-effectors / numbers of robots / type of control, it'll cover those cases cleanly without needing a bunch of boilerplate. I think we're on the right track to do that.

exhaustin:
> 

Yes let's discuss details over VC tmr. Meanwhile some notes in response to the comments:
- I am experimenting around to see how much of the metadata I can include within the urdf file. One concern is that we're currently using the same urdf file for hw and sim, but they require very different gains, at least in the case of a Franka Panda in Pybullet.
- Moving to a in-process controller manager is definitely ideal for several reasons, but it is a relatively complicated task, and if UX is the only immediate gain it's hard to put time into it. However, as I'm trying to finalize the sim API I am keeping in mind that in-process servers are where we want to be ultimately, and ensure that the API does not break when we get there.
- Regarding distinction between arm & gripper, I am leaning heavily on the idea providing control modes named CONTROLLER and DIRECT instead. The former requires the user to send a controller for the server to run, which outputs a vector action space. The latter allows users to directly send the vector action space. Both are supersets of the current arm & gripper interfaces respectively, and seems to be a much more general and clean interface.

polymetis/polymetis/python/polysim/envs/experimental/bullet_manipulator.py
exhaustin: Good catch! I should be using the time in sim instead of the wall clock time in the OS.

polymetis/polymetis/python/polysim/envs/experimental/bullet_manipulator.py
exhaustin: The field exists to identify which of the numerous things that can go wrong on hardware. In the case of sim I'm just setting it to 0. Can probably omit this line since the protobuf message would treat an unassigned int field as 0.

polymetis/polymetis/python/polysim/sim_interface.py
exhaustin: Oops this is leftover code from the previous design iteration. Removing it.

polymetis/polymetis/python/polysim/envs/experimental/bullet_manipulator.py
exhaustin: Oops

polymetis/polymetis/python/scripts/launch_robot.py
exhaustin: `launch_robot.py` was originally written to launch the server in a subprocess using `subprocess.run()` back when a client was always needed. I wanted to make minimal changes to existing files so I had the parent process hang, but will make it more elegant when the design of this PR is finalized.

1heart (commented):
Changes since last review look fine to me.

Minor comment: the several `timestamp.FromNanoseconds(int(1e9 * self.t))` occurrences will result in a lot of data in Jan 1, 1970. Not sure if we want sim envs to have a wall-clock timestamp vs. internal timestamp.

exhaustin:
> Changes since last review look fine to me.
> 
> Minor comment: the several `timestamp.FromNanoseconds(int(1e9 * self.t))` occurrences will result in a lot of data in Jan 1, 1970. Not sure if we want sim envs to have a wall-clock timestamp vs. internal timestamp.

This is a very good point, reverting back to wallclock time as before until we think of a better solution to include sim time in the robot states.

1heart (approved):
Changes since last review lgtm!

PR #1244, opened 2022-06-20
get type name from item stack if it has it

aszlam:
# Description
when creating an ItemStackNode, uses the type name from the struct if the struct has it over the bid meta.


ys1382fb (approved):
image

PR #1260, opened 2022-06-24
Job info manage utils

mialsy:
# Description

Added a util for recording info to record statistics generated during an HITL pipeline run. 

The util class is extendable - more statistic type can be easily added in the future to this class (please refer to the python file documentation for steps to add more statistic types.

Note: A few statistics related to local mephisto db in current hitl pipeline are not updated in this PR, will create new PR for adding those statistics in the future. 

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

After running a nsp pipeline, a json file would be generated that looks like the following:
Screen Shot 2022-06-24 at 3 48 23 PM

# Testing

Manually tested.

# Checklist:


snyxan:
On a high level:
1. We should use some other name than `JobManagementUtil` because it looks a bit confusing. Maybe use something like `Recorder` or `Logger`.
2. We should provide a cleaner abstraction of the `JobManagementUtil`. It should have no prior knowledge of any pipeline/job specific metadata/info to enable better extensibility. All the job specific metadata/info should be defined later when user create pipeline/job specific `JobManagementUtil`, and we should allow users to do so.
3. It seems like `MetaData` can be simplified as RunnerStats, which is on the same level as JobStat. In this way you don't have to distinguish between MetaData and JobStat, and then all set_meta_data & set_job_stat can be unified as something like `record(JOB_TYPE, KEY, VALUE, FORCE_UPDATE)`. This is open to discussion and let me know your thoughts.

droidlet/tools/hitl/utils/job_management.py
```diff
@@ -0,0 +1,296 @@
...
+        repositoryName=AWS_ECR_REPO_NAME,
+        imageIds=[
+            {"imageTag": image_tag},
+        ],
+    )
+    assert len(response["images"]) == 1
+    return response["images"][0]["imageId"]["imageDigest"]
+
+
+def get_s3_link(batch_id: int):
```
snyxan, line 147: This file should only contains the pure abstraction class here. Functions like `get_s3_link` should be moved to a separate util file.

droidlet/tools/hitl/utils/job_management.py
```diff
@@ -0,0 +1,296 @@
...
+            JobStat.MODEL_LOSS,
+        ]
+    ),
+}
+
+# update job stat interval in seconds
+DEFAULT_STAT_UPDATE_INTERVAL = 60
+
+
+def get_dashboard_version(image_tag: str):
```
snyxan, line 132: same as below

PR #1348, opened 2022-08-18
Mirror State Services

laikhtewari:
# Description

Created get/set mirror robot state services. The user client will source robot states from the HW CM and forward them to the local CM using the mirror state setter. The local mirror simulator will pull the mirror states and then update the simulation environment appropriately. While indirect, creating a CM intermediary will facilitate forward simulation (https://www.internalfb.com/tasks/?t=123146501)

Fixes # (issue)

https://www.internalfb.com/tasks/?t=122948322

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

# Testing

TestMirrorRobotState was added to the test_server.cpp file. The cpp tests can be run (after building) with `python ./polymetis/polymetis/tests/cpp/test_server.py`

# Checklist:

### DIFF
--- a/polymetis/polymetis/src/polymetis_server.cpp
+++ b/polymetis/polymetis/src/polymetis_server.cpp
@@ -88,18 +88,12 @@ Status PolymetisControllerServerImpl::GetSimRobotState(
     return Status(StatusCode::FAILED_PRECONDITION,
                   "GetSimRobotState may only be used with a simulator client!");
   }
-  while (true) {
-    if (context->IsCancelled()) {
-      return Status::CANCELLED;
-    }
-    if (is_sim_state_set_) {
-      break;
-    }
-    usleep(SPIN_INTERVAL_USEC);
+  if (!is_sim_state_set_) {
+    return Status(StatusCode::FAILED_PRECONDITION,
+                  "Simulator robot state has not been set!");
   }
   std::lock_guard<std::mutex> guard(sim_state_mtx_);
   robot_state->CopyFrom(sim_robot_state_);
-  is_sim_state_set_ = false;
   return Status::OK;
 }
 
--- a/polymetis/polymetis/src/polymetis_server.cpp
+++ b/polymetis/polymetis/src/polymetis_server.cpp
@@ -70,6 +70,52 @@ Status PolymetisControllerServerImpl::GetRobotStateLog(
   return Status::OK;
 }
 
+Status PolymetisControllerServerImpl::SetMirrorRobotState(
+    ServerContext *context, const RobotState *robot_state, Empty *) {
+  if (!robot_client_context_.metadata.is_sim()) {
+    return Status(
+        StatusCode::FAILED_PRECONDITION,
+        "SetMirrorRobotState may only be used with a simulator client!");
+  }
+  std::lock_guard<std::mutex> guard(mirror_state_mtx_);
+  auto curr_sec = mirror_sim_robot_state_.timestamp().seconds();
+  auto curr_nano = mirror_sim_robot_state_.timestamp().nanos();
+  auto rs_sec = robot_state->timestamp().seconds();
+  auto rs_nano = robot_state->timestamp().nanos();
+  if (!is_mirror_state_set_ || (curr_sec == rs_sec && curr_nano < rs_nano) ||
+      curr_sec < rs_sec) {
+    // update if fresh
+    mirror_sim_robot_state_.CopyFrom(*robot_state);
+    is_mirror_state_set_ = true;
+  }
+  return Status::OK;
+}
+
+Status PolymetisControllerServerImpl::GetMirrorRobotState(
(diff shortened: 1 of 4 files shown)


polymetis/polymetis/src/polymetis_server.cpp
```diff
@@ -70,6 +70,52 @@ Status PolymetisControllerServerImpl::GetRobotStateLog(
...
+    if (context->IsCancelled()) {
+      return Status::CANCELLED;
+    }
+    mirror_state_mtx_.lock();
+    if (is_mirror_state_set_) {
+      break;
+    }
+    mirror_state_mtx_.unlock();
+    usleep(SPIN_INTERVAL_USEC);
+  }
```
exhaustin, line 111: I don't think the locking needs to be this complicated.
In your use case, `SetMirrorRobotState` & `GetMirrorRobotState` are being called from the same thread, and thus there should be no reason for `GetMirrorRobotState` to hang waiting for `SetMirrorRobotState` to complete.
I would simply put lock guards on both functions to prevent getting when the robot state variable is halfway set, but even that isn't needed when your mirror sim client is running within the user process.

On a side note, most low-level C++ objects are thread-safe, so there's no need to put a lock around operations such as checking if a boolean variable is true.

polymetis/polymetis/src/polymetis_server.cpp
```diff
@@ -70,6 +70,52 @@ Status PolymetisControllerServerImpl::GetRobotStateLog(
...
+        StatusCode::FAILED_PRECONDITION,
+        "SetMirrorRobotState may only be used with a simulator client!");
+  }
+  std::lock_guard<std::mutex> guard(mirror_state_mtx_);
+  auto curr_sec = mirror_sim_robot_state_.timestamp().seconds();
+  auto curr_nano = mirror_sim_robot_state_.timestamp().nanos();
+  auto rs_sec = robot_state->timestamp().seconds();
+  auto rs_nano = robot_state->timestamp().nanos();
+  if (!is_mirror_state_set_ || (curr_sec == rs_sec && curr_nano < rs_nano) ||
+      curr_sec < rs_sec) {
```
exhaustin, line 86: You can simply do [`Timestamp.Equals`](https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/well-known-types/timestamp#class_google_1_1_protobuf_1_1_well_known_types_1_1_timestamp_1a00839bf69f69d66efc0bfdd873557c7c).

exhaustin: Honestly I wouldn't even check the timestamp on the server side. The user client can perform this check and thus avoid wasting a request. The server can simply update `is_mirror_state_set_` as the time & computational effort it takes to perform the assignment is trivial compared to sending and receiving the request.

exhaustin (changes requested):
Direction looks perfect but too complicated in general! See comments for details.

polymetis/polymetis/src/polymetis_server.cpp
1heart: Yeah, this is true -- if we are assuming a single producer which only calls SetMirrorRobotState as blocking RPCs, then the requests are never going to be out of order.

polymetis/polymetis/src/polymetis_server.cpp
1heart: I think SetMirrorRobotState is being called from the user client, while GetMirrorRobotState is called from the robot client (two different processes). It's a single-producer 1000Hz, single-consumer 240Hz, single-element-queue situation, which I believe does require synchronization since the consumer needs to pop (both the boolean and the robot state at the same time).

exhaustin: 1. Will the user client be streaming the robot states at 1000hz, and updating the sim at 240hz? I thought sampling by calling `real_robot.GetRobotState` at fixed intervals should be enough for mirroring.
2. I originally thought the mirror sim client would be in a separate process, too. However, the initial implementation in this PR involves having the mirror sim client within the same process as the user client, and after thinking about it a bit I feel like this is the path of least resistance. Should the three of us find some time to discuss this tomorrow?

1heart: I think we do want to also do part 2, where we can send policies to the simulation, not just states, which would require a separate sim process connected to its own controller manager. In that case, the robot client would need to request the robot states at 240Hz in a different process as the user setting the states (at an arbitrary frequency, not necessarily 1kHz).

Added you to a meeting with Laikh this afternoon to sync!

1heart: > having the mirror sim client

Actually, I misunderstood -- this seems reasonable to me, so agreed!

1heart:
Unnecessary due to architectural design change, see #1353 for followup

Issue #1349, opened 2022-08-18
Create RoBoPen Tracking Example


eugeniafeng:
## Type of Issue

Select the type of issue:
- [ ] Bug report (to report a bug)
- [ ] Feature request (to request an additional feature)
- [x] Tracker (I am just using this as a tracker)
- [ ] Refactor request
- [ ] Documentation Ask

## Description

Create Python scripts to run joint/EE trajectory tracking analysis using the RoBoPen dashboard. One for a single joint, performing sinusoidal movement and a chirp function, and another for the end effector following a square.

## Expected Behavior

Joint moves in a sinusoidal motion, and the dashboard displays the difference between the desired and actual position.
End effector moves in a square, and the dashboard displays the difference between the desired and actual position.

PR #1350, opened 2022-08-18
Trajectory tracking example with RoBoPen dashboard

eugeniafeng:
# Description

This PR creates example scripts to run polymetis with the RoBoPen Dashboard. There is one script to move a single joint in a sinusoidal motion, and another to move the end effector in the square. Both will be displayed in the RoBoPen dashboard to allow the user to see the error between desired and actual position or pose. 

Fixes # [(issue)](https://github.com/facebookresearch/fairo/issues/1349)

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

[Chirp function example](https://drive.google.com/file/d/1Hod29psRe3HtbOKk_t5NIE-1gS56Z6Su/view?usp=sharing)
[Square example](https://drive.google.com/file/d/1wuFmh0MjPapA8B52RwyPXifPTMVHnzOK/view?usp=sharing)

# Testing

Follow the instructions in the read me to manually test. This should result in a dashboard similar to the videos above.

# Checklist:


polymetis/polymetis/python/scripts/robopen_dashboard_demos/sinusoidal_movement.py
```diff
@@ -0,0 +1,68 @@
...
+        start = time.time()
+
+        # sinusoidal function
+        joint_positions[target_joint] = (
+            np.sin(np.pi * i / (T * control_hz)) * joint_limit
+        )
+
+        # chirp function
+        # joint_positions[target_joint] = np.sin(np.pi * i * i / (T * control_hz * chirp)) * joint_limit
+
```
exhaustin, line 38: Can you either [define a constant](https://www.pythontutorial.net/python-basics/python-constants/) outside of `__main__` or create an argument using something like [argparse](https://docs.python.org/3/library/argparse.html) to switch between fixed-frequency sinusoidal and chirp? Either one works and is better than commenting out code.

polymetis/polymetis/python/scripts/robopen_dashboard_demos/square_movement.py
```diff
@@ -0,0 +1,110 @@
...
+    #     else:
+    #         ee_pos += torch.Tensor([0.0, 0.0, -side_length])
+
+    #     robot.move_to_ee_pose(ee_pos)
+
+    # Square using update_desired_ee_pose
+    control_hz = 50
+    time_left = 10
+    steps_per_side = control_hz * time_left / 4
+    ee_step = side_length / steps_per_side
```
exhaustin, line 45: Same comment regarding commented out code as above.

exhaustin (changes requested):
Great job wrapping the demo up, code looks clean in general and the thanks for adding the readme! Few minor comments along with the following tasks:
1. The capnp msg files are actually within [fairomsg](https://github.com/facebookresearch/fairo/tree/main/msg/src/fairomsg), so can you update the instructions to pip install from there? This is exactly why we're using a monorepo, so common modules can be shared across projects :)
2. Since the changes are not just a few scripts as I originally anticipated, and are documented and cleaned MUCH better than I was expecting, can you move the entire demo into `fairo/polymetis/examples`? That way you can also revert your fix for having other artifacts in the `scripts` folder. (Awesome job debugging it so quickly, though!)

polymetis/polymetis/python/scripts/robopen_dashboard_demos/sinusoidal_movement.py
eugeniafeng: Yep, fixed! Thanks!

polymetis/polymetis/python/scripts/robopen_dashboard_demos/square_movement.py
eugeniafeng: Thanks! I decided instead to just remove it since it creates a square that pauses on the corners and this section doesn't include pubbing the messages to the dashboard.

exhaustin (approved):
Looks good to merge once tests pass!

PR #1351, opened 2022-08-19
mrp optional process list in import

lshamis:
# Description

Adds an optional parameter to `import_msetup`, allowing users to enumerate a subset of the processes.
```
mrp.import_msetup(path, processes=["proc1", "proc2"])
```

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

# Testing

# Checklist:

### DIFF
--- a/mrp/src/mrp/__init__.py
+++ b/mrp/src/mrp/__init__.py
@@ -1,10 +1,12 @@
-from mrp.process_def import process
+import mrp
+from mrp.process_def import defined_processes, process
 from mrp.runtime.conda import Conda
 from mrp.runtime.docker import Docker
 from mrp.runtime.host import Host
 from mrp.util import NoEscape
 from importlib.machinery import SourceFileLoader
 import click
+import contextlib
 import inspect
 import os
 import sys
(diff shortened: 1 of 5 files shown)


mrp/src/mrp/__init__.py
```diff
@@ -24,17 +25,36 @@ def main(*args):
...
+
+    # Attach the hook.
+    mrp.process = hook
+
+    # Import.
     SourceFileLoader("msetup", target_path).load_module()
 
+    # Undo the hook.
+    mrp.process = process_impl
+
```
exhaustin, line 57: 1. Curious if there's a reason to go through the relatively opaque method of hooking `mrp.process`. Why not simply do something like `loader = SourceFileLoader("msetup", target_path); [loader.load_module(name) for name in processes]`?
2. I noticed [`SourceFileLoader.load_module`](https://docs.python.org/3/library/importlib.html#importlib.machinery.SourceFileLoader.load_module) is being deprecated, is that something of concern?

exhaustin (approved):
One small comment, otherwise looks good!

mrp/src/mrp/__init__.py
lshamis: I don't think the string passed to `load_module` is the name of the symbol to import. Even if it was, processes aren't Python symbols. They are entries in the `defined_processes` table.

PR #1355, opened 2022-08-24
forward sim + refactor

laikhtewari:
# Description

The user interface was abstracted such that a user can run forward simulation on a local mirror sim by simply turning on a flag. 

Forward sim was implemented by launching the mirror sim client run method in a background thread while the user is in forward sim mode.

Fixes # (issue)
[T123146501](https://www.internalfb.com/intern/tasks/?t=123146501)
## Type of change

- [x] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

Before: 
    sim_robot = RobotInterface(ip_address=sim_ip, port=sim_port)
    hw_robot = RobotInterface(ip_address=hw_ip, port=hw_port)

    hw_robot.go_home()
    mirror_sim_client.sync(hw_robot)  # must be non-blocking
    hw_robot.send_torch_policy(policy)
    mirror_sim_client.unsync()

After:
    hw_robot = RobotInterface(
        ip_address=hw_ip,
        port=hw_port,
        use_mirror_sim=True,
        mirror_cfg=cfg,
        mirror_ip=sim_ip,
        mirror_port=sim_port,
    )
    hw_robot.setup_mirror_for_forward()
    hw_robot.go_home(use_mirror=True)
    hw_robot.send_torch_policy(policy, use_mirror=True)
    hw_robot.clean_mirror_after_forward()

    hw_robot.sync_with_mirror()
    hw_robot.go_home()
    hw_robot.send_torch_policy(policy)
    hw_robot.unsync_with_mirror()

# Testing

# Checklist:


polymetis/polymetis/python/polymetis/robot_interface.py
```diff
@@ -744,3 +784,32 @@ def move_ee_xyz(
...
+        assert (
+            not self.mirror_sim_client._state_setter
+        ), "Unsync before setting up for forward!"
+        self.mirror_sim_client.run(time_horizon=1)
+        self.mirror_sim_client.run_no_wait()
+        self.mirror_sim_robot = RobotInterface(
+            ip_address=self.mirror_ip,
+            port=self.mirror_port,
+            mirror_metadata=self.mirror_sim_client.metadata.get_proto(),
+        )
```
exhaustin, line 810: Would it be easier to simply perform this by default in `__init__` if `use_mirror_sim == True`?

polymetis/polymetis/python/polymetis/robot_interface.py
```diff
@@ -744,3 +784,32 @@ def move_ee_xyz(
...
+    # MIRROR METHODS
+
+    def sync_with_mirror(self):
+        assert self.use_mirror_sim, "Mirror sim must be instantiated!"
+        assert not self.mirror_sim_robot, "Clean mirror after forward in order to sync!"
+        self.mirror_sim_client.sync(self)
+
+    def unsync_with_mirror(self):
+        assert self.use_mirror_sim, "Mirror sim must be instantiated!"
+        self.mirror_sim_client.unsync()
```
exhaustin, line 797: The user doesn't really need access to these methods. We should be able to unsync before running any policy and resync afterwards within `RobotInterface`.

polymetis/scripts/mirror_sim_tester.py
```diff
@@ -0,0 +1,47 @@
...
+        Kqd=Kqd_default,
+        Kx=Kx_default,
+        Kxd=Kxd_default,
+        robot_model=robot.robot_model,
+        ignore_gravity=robot.use_grav_comp,
+    )
+
+    # Run policy
+    print("\nRunning custom policy ...\n")
+    state_log = robot.send_torch_policy(policy)
```
exhaustin, line 47: No need to have the test script within the PR, unless you want to polish it into an integration test or an example.

polymetis/scripts/mirror_sim.py
```diff
@@ -61,19 +65,23 @@ def main(cfg):
...
     # Mirror
+    print("Syncing...")
+    hw_robot.sync_with_mirror()
     print("Homing robot...")
     hw_robot.go_home()
-    print("Syncing...")
-    mirror_sim_client.sync(hw_robot)  # must be non-blocking
     print("Sending policy...")
     hw_robot.send_torch_policy(policy)
     print("Unsyncing...")
```
exhaustin, line 83: This still seems quite complicated from the user side.
Can the user simply do 
```py
robot.go_home()   # mirror_sim also goes home automatically because they're in sync
robot.send_torch_policy(policy, use_mirror=True)
robot.send_torch_policy(policy)
```

polymetis/polymetis/python/polymetis/robot_interface.py
```diff
@@ -197,6 +208,13 @@ def send_torch_policy(
             If `blocking`, returns a list of RobotState objects. Otherwise, returns None.
 
         """
+        if use_mirror:
+            assert (
+                self.mirror_sim_robot
+            ), "Must call setup_mirror_for_forward before forward methods can be used on mirror sim"
+            return self.mirror_sim_robot.send_torch_policy(
+                torch_policy, blocking, timeout, use_mirror=False
+            )
```
exhaustin, line 217: This block of code is repeated in `send_torch_policy` and `go_home`, and probably will also be in `move_to_joint_positions` and `move_to_ee_pose` in the future. Can you make a private method to deduplicate?

polymetis/polymetis/python/polymetis/robot_interface.py
```diff
@@ -285,6 +303,10 @@ def __init__(
         self,
         time_to_go_default: float = 1.0,
         use_grav_comp: bool = True,
+        use_mirror_sim: bool = False,
+        mirror_cfg: DictConfig = None,
+        mirror_ip: str = "",
+        mirror_port: int = -1,
```
exhaustin, line 309: In the current implementation, configuration of the mirror sim is spread across both the parent class (`BaseRobotInterface`) and the child class (`RobotInterface`) without an obvious logic to which part should be in which class.
I think it would be cleaner to either:
1. Set up all mirror sim member variables and methods in parent class (`BaseRobotInterface`). The child class will simply be calling the appropriate methods in `go_home` (and maybe `move_to_joint_positions` and `move_to_ee_pose`)
2. Set up all mirror sim member variables and methods in child class (`RobotInterface`). You would have to call `super().send_torch_policy` within `RobotInterface.send_torch_policy` in order to add mirror sim stuff to the inherited method.

Let me know if you have any other thoughts on the design, happy to iterate!

exhaustin (commented):
Direction looks correct and looks like all the core functionalities are implemented quite well!
Remaining subtasks are mostly code cleanup, API polishing, and testing.

polymetis/polymetis/python/polymetis/robot_interface.py
exhaustin: Decorators are probably an elegant solution to this, if you're familiar with them.

laikhtewari:
Tested forward simulation and mirroring on franka panda hardware robot with @1heart and observed appropriate robot states

polymetis/scripts/mirror_sim.py
laikhtewari: Defaulting to mirror mode and switching to forward mode only when running a policy on the mirror sim would enable this simpler UI. I think only blocking send policy calls would be supported to prevent concurrently mirroring and running a forward policy. Does this sound good @exhaustin?

exhaustin: I think this makes sense. I don't think there are any users that are using the non-blocking version of the move_to calls, @1heart can you confirm?

1heart: What about everyone using the `start_*_impedance`, `update_desired_*_positions` methods? Those are non-blocking.

Maybe a cleaner way to do this is with the `with` paradigm:


```
with hw_robot.viz_mirror():
   hw_robot.go_home()  # hw reflected in sim

with hw_robot.viz_sim():  # or some other name for this
    hw_robot.send_torch_policy(policy)  # only reflected in sim
```

1heart (approved):

PR #1362, opened 2022-08-28
Mob pick

aszlam:
# Description
allow mobs to pick things up in pyworld.  also fix a small clarification setup bug


jacklanchantin (approved):

jacklanchantin (commented):
If an object isn't in the agent's possession it's always set as ["in_others_inventory"](https://github.com/facebookresearch/fairo/blob/be8cc33ab5b15ecff2c0d4e95956ce782d05034b/droidlet/memory/craftassist/mc_memory.py#L194). Should we specify who has it there?

Issue #1398, opened 2022-11-11
Gripper stuck when grasping object


JonathanYang0127:
## Type of Issue

Select the type of issue:
- [x] Bug report (to report a bug)
- [x] Feature request (to request an additional feature)
- [ ] Tracker (I am just using this as a tracker)
- [ ] Refactor request
- [ ] Documentation Ask

## Description

When using the gripper goto function (regardless of blocking), the franka hand gripper gets stuck when asked to fully close when there is an object between its fingers. As a result, it does not respond to all future commands because it continuously attempts to reach the fully closed position.

## Expected Behavior

With blocking off, the gripper should respond to a new command regardless of whether the old one was achieved.

AlexanderKhazatsky:
I'm experiencing this issue as well. Specifically, I'm experiencing two issues:

1) The gripper pauses unnecessarily when going to a new width (ex from fully open to fully closed)
2) When the gripper tries to grasp an object, if it isn't able to fully able to go to the desired location, it gets stuck - even after I update the desired state to be fully open. It only starts to move after I remove the object blocking the gripper fingers from closing.

AlexanderKhazatsky:
Here’s some videos of the issues discussed:


https://user-images.githubusercontent.com/30299984/201435469-aa006e6f-0fdd-457d-9e85-42ada36fbd30.MOV

AlexanderKhazatsky:
https://user-images.githubusercontent.com/30299984/201436229-9497d19c-6e17-46a0-bcf7-3f93b7e2ba3f.mov

AlexanderKhazatsky:
Wanted to bump this issue :)

exhaustin:
Hello sorry for the slow response. I cannot change the behavior of `goto` for the Franka Hand as we are simply wrapping around [franka::Gripper::move](http://docs.ros.org/en/kinetic/api/libfranka/html/classfranka_1_1Gripper.html#a331720c9e26f23a5fa3de1e171b1a684).
What I can do is to expose [`franka::Gripper::grasp`](http://docs.ros.org/en/kinetic/api/libfranka/html/classfranka_1_1Gripper.html#a19b711cc7eb4cb560d1c52f0864fdc0d), which is designed to avoid this issue when grasping. I've done that in the PR mentioned above and will close the issue since there is nothing I can do more than that.

[...]

PR #1399, opened 2022-12-02
[Polymetis] Completely expose Franka Hand gripper API

exhaustin:
# Description

Issue #1398 highlights an inflexibility of [`franka::Gripper::move`](http://docs.ros.org/en/kinetic/api/libfranka/html/classfranka_1_1Gripper.html#a331720c9e26f23a5fa3de1e171b1a684), which is currently the only API Polymetis is using to command the Franka Hand.
Basically the move command blocks until it reaches the desired goal, which causes problems during grasping.
This PR exposes [`franka::Gripper::grasp`](http://docs.ros.org/en/kinetic/api/libfranka/html/classfranka_1_1Gripper.html#a19b711cc7eb4cb560d1c52f0864fdc0d), which is a clunky API that at least allows the user to execute a grasp instead of commanding the gripper to 0 width. 

## Type of change

- [ ] Refactor
- [ ] This is a unit test
- [ ] Datasets Release
- [ ] Models Release

## Type of requested review

## Before and After

# Testing

Tested on hardware

# Checklist:

### DIFF
--- a/polymetis/polymetis/include/polymetis/clients/franka_hand_client.hpp
+++ b/polymetis/polymetis/include/polymetis/clients/franka_hand_client.hpp
@@ -25,6 +25,7 @@ private:
   GripperState gripper_state_;
   GripperCommand gripper_cmd_;
   int prev_cmd_timestamp_ns_;
+  bool prev_cmd_successful_ = true;
 
   // Franka
   std::shared_ptr<franka::Gripper> gripper_;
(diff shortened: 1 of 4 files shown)


1heart (approved):
Makes sense.

One question -- it looks like epsilon_inner/outer have different default values in Python/C++. Where are the values coming from?

exhaustin:
> Makes sense.
> 
> One question -- it looks like epsilon_inner/outer have different default values in Python/C++. Where are the values coming from?

That was from my original misunderstanding of how these parameters work. I thought they were upper and lower limits, and so I set the `epsilon_inner` to slightly below zero and `epsilon_upper` to larger than max width.
After learning that they were tolerances from the target grasp width, I just default them to large numbers.

1heart:
Seems like mrp tests are broken here? Not sure why

